PR #9, opened 2016-08-14
add --state flag and pass it to swupd

daniel-schuermann:
Avoid creating the swupd state within the image creation directory by providing a -S/--state flag and passing it to swupd.
Patch created by Leoandro Dorileo.
Added unit tests.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -380,6 +380,8 @@ def copy_os(args, template, target_dir):
     swupd_command += " --contenturl={0}".format(args.contenturl)
     swupd_command += " --versionurl={0}".format(args.versionurl)
     swupd_command += " --format={0}".format(args.format)
+    swupd_command += " --statedir={0}".format(args.statedir)
+
     if template["DestinationType"] == "physical":
         os.makedirs("/var/lib/swupd", exist_ok=True)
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -380,6 +380,10 @@ def copy_os(args, template, target_dir):
     swupd_command += " --contenturl={0}".format(args.contenturl)
     swupd_command += " --versionurl={0}".format(args.versionurl)
     swupd_command += " --format={0}".format(args.format)
+
+    if args.state is not None:
```
bryteise: So looking at this a little closer I should have reviewed the patch I sent you better. I think instead of None, args.state should default to '/var/lib/swupd'. The reason for this is to fit in with the rest of the options passed for contenturl, versionurl and format better.

ister.py
```diff
@@ -1243,6 +1247,9 @@ def handle_options():
     parser.add_argument("-k", "--kcmdline", action="store",
                         default="/proc/cmdline",
                         help="File to inspect for kernel cmdline opts")
+    parser.add_argument("-S", "--state", action="store",
```
bryteise: Also, instead of state being the argument statedir would be a more fitting name.

PR #10, opened 2016-08-17
Ensure trademark/branding compliance in ister_gui

matthewrsj:
- First mention needs to be "Clear Linux OS for Intel Architecture"
- Any later mentions can be short form, "Clear Linux OS"
- Avoid any other spellings or mentions other than these two


bryteise:
Rebased and merged.

PR #11, opened 2016-08-17
Suggested username contains only valid charactars

matthewrsj:
Users with spaces in their last names were experiencing a bug where the
suggested username (created using their last name) included spaces.
This is fixed by checking each character before adding it to the
suggested username.

The following regular expression is used to create the suggestion:

	[a-z_][a-z0-9-_]*[$]


bryteise:
merged

PR #12, opened 2016-08-17
Order ister gui options

matthewrsj:
The options are defined in ister_gui as python dictionaries, which
are inherently unordered. By calling sorted() on the options the
dictionaries are guaranteed to be sorted by key.


bryteise:
merged

PR #13, opened 2016-08-18
Change first two screens to real screens

matthewrsj:
The installer process should be fully navigable front-to-back. Network
splash screen and telemetry opt-in screen now implemented as fully
navigable screens.


bryteise:
merged.

PR #17, opened 2016-08-19
Simplify keypress navigation

matthewrsj:
Remove the ability to navigate 'up' or 'shift tab' between widgets.
These keys are handled by urwid in a way that causes our focus to lose
track of its location. This causes the focus to jump in unexpected and
unpredictable ways. 'shift tab' and 'up' still work within widgets, such
as the FormBody and NavBar widgets. 'tab' and 'right arrow' still work
in the expected way, allowing the user to navigate between widgets.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -272,9 +272,6 @@ class ButtonMenu(object):
         if 'tab' in keys or 'down' in keys:
             self._lb.focus_position = (i + 1) % max_pos
             self._lb._invalidate()
-        elif 'shift tab' in keys or 'up' in keys:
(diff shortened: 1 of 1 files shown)


bryteise:
Is down the universal position here or does right make sense or get thrown off by urwid too?

matthewrsj:
Down is universal. I don't think the right arrow should work in the form body. I've seen some GUIs that use right to change configuration options. The right arrow works in the navbar though.

bryteise:
Ah okay, wasn't sure if the navbar was included in this. Merged.

PR #18, opened 2016-08-20
Provide progress feedback to user

matthewrsj:
Provide feedback in the form of 'Step x of n' where n can grow depending
on the user's choices during the install process.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1333,7 +1333,7 @@ class ConfirmUserMenu(ProcessStep):
         raise urwid.ExitMainLoop()
 
     def build_ui(self):
-        self._ui = SimpleForm(u'Choose user configuration', self._ui_widgets,
+        self._ui = SimpleForm(u'User configuration', self._ui_widgets,
                               buttons=["Previous"])
 
     def run_ui(self):
@@ -1482,7 +1482,7 @@ class ConfirmDHCPMenu(ProcessStep):
         raise urwid.ExitMainLoop()
 
     def build_ui(self):
-        self._ui = SimpleForm(u'Choose network configuration',
+        self._ui = SimpleForm(u'Network configuration',
                               self._ui_widgets,
                               buttons=["Previous"])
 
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -1653,22 +1679,24 @@ def __init__(self, *args, **kwargs):
...
-        confirm_dhcp = ConfirmDHCPMenu()
-        static_ip_config = StaticIpStep()
+        set_mount_points = MountPointsStep(4, 9)
+        bundle_selector = BundleSelectorStep(6, 9)
+        confirm_user = ConfirmUserMenu(7, 9)
+        user_configuration = UserConfigurationStep(8, 10)
+        confirm_dhcp = ConfirmDHCPMenu(8, 9)
+        confirm_dhcp2 = ConfirmDHCPMenu(9, 10)
+        static_ip_config = StaticIpStep(9, 10)
+        static_ip_config2 = StaticIpStep(10, 11)
```
bryteise: So the numbering here seems a little funny since there is never a step 9 of 9 or 11 of 11.

ister_gui.py
```diff
@@ -1705,12 +1733,17 @@ def _init_actions(self):
...
         confirm_dhcp.set_action('dhcp', confirm_installation)
         confirm_dhcp.set_action('staticip', static_ip_config)
         confirm_dhcp.set_action('Previous', confirm_user)
+        confirm_dhcp2.set_action('dhcp', confirm_installation)
+        confirm_dhcp2.set_action('staticip', static_ip_config2)
+        confirm_dhcp2.set_action('Previous', confirm_user)
         static_ip_config.set_action('Previous', confirm_dhcp)
         static_ip_config.set_action('Next', confirm_installation)
+        static_ip_config2.set_action('Previous', confirm_dhcp)
+        static_ip_config2.set_action('Next', confirm_installation)
```
bryteise: These don't seem quite the same as the other changes but I'm guessing they are here to get the numbering right? Normally this would be part 1 of a 2 part patch series I think.

ister_gui.py
matthewrsj: 9/9 and 11/11 is the `ConfirmStep()` which tells the user the setup is complete and asks the user to confirm if they want to begin the install. I assumed the user would be able to figure out it was the last step.

ister_gui.py
matthewrsj: The installer is not linear, so this method can be called in a different step x of n. I have to have two instances here so call the same method with different arguments.

ister_gui.py
bryteise: Hrm well if we don't say that is a counted step maybe it should be 8/8 and 10/10 then?

PR #19, opened 2016-08-22
Add --dry-run argument option

matthewrsj:
When ister_gui.py exits, it calls either `reboot` or `poweroff`, even
when it is run locally with `sudo python ister_gui.py`. Unless
interrupted with CTRL-C, the installer shuts down the host on exit.
This adds a command line argument (-l, --local-test) that specifies
that no install should run and the system should not reboot or
poweroff on exit.


bryteise:
Happy with the implementation but usually this is option is '--dry-run'.

bryteise:
Merged.

PR #22, opened 2016-08-25
Add '-- required --' text above required bundles

matthewrsj:
Without the text divider, the bundles look like it is possible to
select and unselect them. When tabbing through, the focus actually
skips over the required bundles, causing the user to blow past the
'Next' button.


bryteise:
Merged in https://github.com/bryteise/ister/commit/f34aa476e302adac9febb5d4138622e34d168bb0.

PR #24, opened 2016-08-30
Add network check on splash screen

matthewrsj:
Display network requirements and give indication of whether
the system is able to connect to clearlinux.org or not. Allow the
user to configure a static ip address, gateway, and interface for the
installer to use.

Also add configurable proxy to installer gui and pass to swupd via the
json template. This way, if pacrunner is not working the user can
define their own proxies to connect to clearlinux.org for the
installation.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -60,7 +60,7 @@ LOG = None
 DEBUG = False
 
 
-def run_command(cmd, raise_exception=True, log_output=True):
+def run_command(cmd, raise_exception=True, log_output=True, environ=None):
     """Execute given command in a subprocess
 
     This function will raise an Exception if the command fails unless
@@ -71,7 +71,8 @@ def run_command(cmd, raise_exception=True, log_output=True):
         sys.stdout.flush()
         proc = subprocess.Popen(cmd.split(" "),
                                 stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
+                                stderr=subprocess.PIPE,
+                                env=environ)
         lines = proc.stdout
         output = []
         for line in lines:
@@ -398,7 +399,14 @@ def copy_os(args, template, target_dir):
         os.makedirs("{0}/var/tmp".format(target_dir))
         run_command("mount --bind {0}/var/tmp /var/lib/swupd"
                     .format(target_dir))
-    run_command(swupd_command)
+    swupd_env = os.environ
+    if "Proxy" in template and template["Proxy"]:
+        proxy = "https_proxy={} ".format(template["Proxy"])
+        swupd_env["http_proxy"] = proxy
+        swupd_env["https_proxy"] = proxy
+        LOG.debug("proxy: {}".format(proxy))
+
+    run_command(swupd_command, environ=swupd_env)
 
 
 class ChrootOpen(object):
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -50,7 +50,10 @@ import subprocess
 import threading
 import sys
 import pprint
+import time
 import ipaddress
+import netifaces
+import pycurl
 
 # pylint: disable=E0401
 import urwid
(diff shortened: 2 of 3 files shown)


ister.py
```diff
@@ -384,8 +385,8 @@ def copy_os(args, template, target_dir):
     if not args.format:
         args.format = get_current_format()
     add_bundles(template, target_dir)
-    swupd_command = "swupd verify --install --path={0} " \
-                    "--manifest={1}".format(target_dir, template["Version"])
```
bryteise: Did this get moved because of the line limit?

ister.py
```diff
@@ -398,7 +399,13 @@ def copy_os(args, template, target_dir):
         os.makedirs("{0}/var/tmp".format(target_dir))
         run_command("mount --bind {0}/var/tmp /var/lib/swupd"
                     .format(target_dir))
-    run_command(swupd_command)
+    swupd_env = os.environ
+    if "Proxy" in template and template["Proxy"]:
+        proxy = "https_proxy={} ".format(template["Proxy"])
+        swupd_env["https_proxy"] = proxy
+        LOG.debug("proxy: {}".format(proxy))
```
bryteise: The update server could also be on an http endpoint as well.

ister.py
matthewrsj: ah, I may have changed that with some code I added there then removed. I can change it back.

ister_gui.py
```diff
@@ -641,24 +643,215 @@ def validate_condition(self):
         return True
 
 
-class SplashScreen(ProcessStep):
+class NetworkRequirements(ProcessStep):
     """UI to select automatic or manual installation"""
```
bryteise: This comment seems out of place with this class name.

ister_gui.py
```diff
@@ -641,24 +643,215 @@ def validate_condition(self):
...
+            return self._action
+
+        self.error = ''
+        try:
+            ipaddress.ip_address(self.static_ip.get_edit_text())
+        except Exception as err:
+            self.error = 'The configuration for "ip" is invalid {}'.format(err)
+            # an empty action signals to refresh the page
+            self._action = ''
+        if self.interface.get_edit_text() not in self.ifaceaddrs.keys():
```
bryteise: Since I haven't run this, is the user able to set configurations for interfaces that don't exist?

matthewrsj: no they are not

ister_gui.py
```diff
@@ -641,24 +643,215 @@ def validate_condition(self):
...
+            config = self.config
+            return self._action
+        else:
+            Alert('Error!', self.error).do_alert()
+            self.error = ''
+            return self._action
+
+    def build_ui_widgets(self):
+        """Build ui handler
+        The last urwid Divider helps the tab movement to work"""
```
bryteise: Maybe I'm off because this is a class method but I thought docstrings were supposed to have an empty line separating the first line from the body.

matthewrsj: I thought it was fine in methods. PEP8 doesn't complain about them. I can add a blank line if you want me too.

ister_gui.py
```diff
@@ -641,24 +643,215 @@ def validate_condition(self):
...
+    def build_ui(self):
+        self._ui = SimpleForm(u'Network Requirements',
+                              self._ui_widgets, buttons=["Refresh", "Ok"])
+
+    def run_ui(self):
+        return self._ui.do_form()
+
+    def _network_connection(self):
+        """Check if connection to https://www.clearlinux.org is available"""
+        cmd = ['/usr/bin/curl', '-s', '-I', 'https://www.clearlinux.org',
```
bryteise: The installation images do not have curl by default, see comment in ister.py on validating network that kind of goes with this. Seems the desire was to use pycurl.

bryteise: I should say they do not have the curl command, they have libcurl however.

matthewrsj: I was trying to avoid pycurl because it added a dependency. The comment in ister.py seems to refer to replacing urllib2

ister_gui.py
```diff
@@ -641,24 +643,215 @@ def validate_condition(self):
...
+            if af_inet in ifaddrs:
+                ip_addrs = ifaddrs[af_inet][0]
+                if ip_addrs['addr'] and '127.0.' not in ip_addrs['addr']:
+                    addrs[if_name] = ip_addrs['addr']
+
+        # pylint: enable=E1103
+        self.ifaceaddrs = addrs
+        for interface in addrs:
+            if interface.startswith('e'):
+                return (interface, addrs[interface])
```
bryteise: So the above adds a lot of non gui functionality to ister_gui.py and I'm worried we won't have unit test coverage for it just the expect scripts. I'm starting to think an ister python library may be the best place to put some of these features unless you want to extend ister_test.py to import the gui for testing. Up to you on what you think would be best.

ister_test.py
```diff
@@ -468,6 +468,11 @@ def run_command_good():
     ister.run_command("true")
 
 
+def run_command_with_env():
+    """run_command with environment variable passed"""
+    ister.run_command("true", environ=os.environ)
```
bryteise: These run command tests are not very good =(. I feel bad about that as they really should be mocked but more to validate correctness but I didn't do that yet. So given that this test is fine but it really doesn't tell you if it is using the environment or not from the test result.

ister_test.py
```diff
@@ -116,7 +116,7 @@ def run_command_wrapper(func):
     def wrapper():
         """run_command_wrapper"""
         def mock_run_command(cmd, _=None, raise_exception=True,
-                             log_output=True):
+                             log_output=True, environ=None):
```
bryteise: So the environ should be added to the COMMAND_RESULTS list when not none in order to validate it is actually getting the expected environ content.

ister_gui.py
bryteise: No that if pep8 doesn't complain I am not going to =).

ister_gui.py
bryteise: So either way we are adding a new dependency to the project because it has to use libcurl in order to use the pacrunner bits.

ister_test.py
```diff
@@ -3800,6 +3801,18 @@ def mock_modify_cloud_init_service_file(target_dir):
     commands_compare_helper(commands)
 
 
+def gui_network_connection_with_proxy():
+    netreq = ister_gui.NetworkRequirements()
+    netreq.config = {"proxy": "https://proxy.to.clearlinux.org:1080"}
+    netreq._network_connection()
```
bryteise: So I'd mock out the pycurl.Curl() to give you a class you just introspec to see that the netreq.config is applied correctly and to not actually perform the request.

ister_test.py
```diff
@@ -3800,6 +3801,18 @@ def mock_modify_cloud_init_service_file(target_dir):
...
+
+
+def gui_network_connection_no_proxy():
+    netreq = ister_gui.NetworkRequirements()
+    netreq.config = {}
+    netreq._network_connection()
```
bryteise: Similarly here for mocking.

ister_test.py
```diff
@@ -55,6 +55,7 @@
 import netifaces
 
 import ister
+import ister_gui
```
bryteise: I want to have the static network detection and static file writing logic tested here most of all I think, wrap the open call with the decorator to mock it.

matthewrsj:
@bryteise I made those changes

matthewrsj:
@bryteise test updated

PR #25, opened 2016-08-31
Set user's full name in /etc/passwd

matthewrsj:
The gui provides capability for the user to set their first and
last name during setup, but the installer did not set the name
in /etc/passwd. This uses the chfn tool to set the user's full
name.

### DIFF
--- a/ister_test.py
+++ b/ister_test.py
@@ -1519,6 +1519,7 @@ def add_users_good():
     backup_add_user_key = ister.add_user_key
     backup_setup_sudo = ister.setup_sudo
     backup_disable_root_login = ister.disable_root_login
+    backup_add_user_fullname = ister.add_user_fullname
 
     def mock_create_account(user, target_dir):
         """mock_create_account wrapper"""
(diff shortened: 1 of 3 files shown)


ister_test.py
```diff
@@ -1589,6 +1597,29 @@ def mock_create_account(_, __):
...
+        COMMAND_RESULTS.extend(cmd)
+
+    backup_call = subprocess.call
+    subprocess.call = mock_call
+
+    template = {"fullname": "Test User", "username": "user"}
+    commands = ["chfn", "-f", "Test User", "user"]
+    ister.add_user_fullname(template, "/tmp")
+    subprocess.call = backup_call
+    if COMMAND_RESULTS != commands:
```
bryteise: Is there a reason to not use the commands_compare_helper here?

matthewrsj: not a good one. fixed

bryteise:
Merged in bd2e6d28e4798503c6849d54324299f8d593f69b.

PR #26, opened 2016-09-01
Improve proxy setting in ister.py

matthewrsj:
In copy_os the proxy was being set in a way that resulted in the
environment proxies formatted like "https_proxy=https_proxy=<proxy>".

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -400,11 +400,10 @@ def copy_os(args, template, target_dir):
         run_command("mount --bind {0}/var/tmp /var/lib/swupd"
                     .format(target_dir))
     swupd_env = os.environ
-    if "Proxy" in template and template["Proxy"]:
-        proxy = "https_proxy={} ".format(template["Proxy"])
-        swupd_env["http_proxy"] = proxy
-        swupd_env["https_proxy"] = proxy
-        LOG.debug("proxy: {}".format(proxy))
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -400,11 +400,10 @@ def copy_os(args, template, target_dir):
...
     swupd_env = os.environ
-    if "Proxy" in template and template["Proxy"]:
-        proxy = "https_proxy={} ".format(template["Proxy"])
-        swupd_env["http_proxy"] = proxy
-        swupd_env["https_proxy"] = proxy
-        LOG.debug("proxy: {}".format(proxy))
+    if template.get("Proxy"):
+        swupd_env["http_proxy"] = template["Proxy"]
+        swupd_env["https_proxy"] = template["Proxy"]
+        LOG.debug("proxy: {}".format(template["Proxy"]))
```
bryteise: Looking at how this didn't get picked up in the test. It may be good to have the run_command_wrapper (https://github.com/bryteise/ister/blob/master/ister_test.py#L115) changed to append the http and https values directly since nothing else from the environment is used at this time.

bryteise:
Merged in https://github.com/bryteise/ister/commit/4853173a453c5cb4ac652cfbb4ed4d17f3a72bef

PR #28, opened 2016-09-08
Add ChooseAction screen for shell/repair/install paths

matthewrsj:
Allows the user to select which path to take:
  - Install leads to install path
  - Repair mounts the host os disk and runs swupd verify --fix in
    its chroot.
  - Shell gives the user a shell in the host os disk chroot
  - Update Installer runs swupd update in the installer image and
    restarts the installer, giving the user some survival options.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -54,6 +54,8 @@ import time
 import ipaddress
 import netifaces
 import pycurl
+import tempfile
+import shutil
 
 # pylint: disable=E0401
 import urwid
@@ -136,6 +138,32 @@ def get_disk_info(disk):
         else:
             break
     return info
+
+
+def get_list_of_disks():
+    """"Queries for the available disks discarding the inst. source"""
+    dir_path = '/sys/block'
+    disks = [device for device in os.listdir(dir_path)
+             if 'pci' in os.readlink('{0}/{1}'.format(dir_path, device))]
+    unmounted = list()
+
+    output = subprocess.check_output([
+        '/usr/bin/lsblk', '-o',
+        'MOUNTPOINT,PARTUUID']).decode('utf-8')
+    parts = output.split('\n')
+
+    expr = re.compile(r'(\S*)\s*(\S*)')
+    for line in parts:
+        match = expr.match(line)
+        if match:
+            if match.group(1) == '/':
+                part_uuid = match.group(2)
+
+    root_disk = os.readlink('/dev/disk/by-partuuid/{0}'.format(part_uuid))
+    for disk in disks:
+        if disk not in root_disk:
+            unmounted.append(disk)
+    return unmounted
 
 
 class Alert(object):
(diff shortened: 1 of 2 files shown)


matthewrsj:
This uses the old method of setting the installer proxy. Please hold on merging this until after #30 has been merged. I will update this to use the new method once that is done.

matthewrsj:
@bryteise tests added and updated to new proxy method. Happy reviewing.

ister_gui.py
```diff
@@ -644,6 +672,247 @@ def validate_condition(self):
...
 
+class ChooseAction(ProcessStep):
+    """UI to select installation path (Install, Shell, or Repair)"""
+    def __init__(self):
+        super(ChooseAction, self).__init__()
+        self.choices = {
+            'Install': 'Install          - Install Clear Linux OS',
+            'Repair': 'Repair           - Attempt to repair host OS',
+            'Shell': 'Shell            - Open shell in host OS',
+            'Update Installer': 'Update Installer - Update installer to '
```
bryteise: This option is unusual and I'm not super excited about bugs where the installer failed after an update to the installer image due to who knows what instead of just grabbing a new installer image and trying that instead. Could use some discussion on why we should add it.

ister_gui.py
```diff
@@ -644,6 +672,247 @@ def validate_condition(self):
...
+            self.repair_host(config)
+
+        if self._action == 'Update Installer':
+            self.update_installer(config)
+
+        return self._action
+
+    def chroot_shell(self, config):
+        """Chroot into host os (if one is found) and provide a shell"""
+        self.error = False
```
bryteise: Shouldn't need to set this here when init also does the same. Makes me think this is handling something else not cleaning up properly.

matthewrsj: init is only run the first time that the user loads this screen. After that handler is run on returns.

ister_gui.py
```diff
@@ -644,6 +672,247 @@ def validate_condition(self):
...
+                term = Terminal(['bash'])
+
+        try:
+            if term:
+                term.main_loop()
+        except Exception as excep:
+            Alert('Error!', 'Error occurred in shell: {}'
+                            .format(excep)).do_alert()
+
+        self._umount_host_disk(target_dir, os_disk)
```
bryteise: Should clean up and reset self.error here I think.

ister_gui.py
```diff
@@ -644,6 +672,247 @@ def validate_condition(self):
...
+        """
+        cmd = ['lsblk', '-l', '-o', 'NAME,MOUNTPOINT']
+        output = subprocess.check_output(cmd).decode('utf-8')
+        for line in output.split('\n'):
+            if '/' in line:
+                return line.split()[0]
+
+        return ''
+
+    def _mount_host_disk(self, config):
```
bryteise: This should mount the boot partition as well if possible.

ister_gui.py
bryteise: Hrm do you ignore some things setting self.error? If you do then setting this before testing makes sense. Otherwise I'd say any path that acts on a self.error should also reset it after it checks.

Prevents cases where:

self.error = False
self.do_one()
if self.error:
 self.do_two()
else:
 self.do_three()
if self.error:
  # are we sure do_two and do_three manage self.error or did only do_two set it?

ister_gui.py
matthewrsj: All this does is run `swupd update` in the installer image. Don't we want our users to always have the most up-to-date installer? If this fails the user can do as you suggest and grab a new installer image. I don't see how updating the image they are on would cause any bugs, unless swupd fails in a strange way.

rnesius: If it goes off the rails they can always fall back to that. Agree it's a bit different - seems to work just fine though.

bryteise: So swupd update on a fixed size installer image that attempts to be small does not give me much confidence. Also those files are cached in /var and cleanup isn't automated at this point.

I'm just not really convinced adding any code for this is worthwhile and fixes more issues than it brings up.

ister_gui.py
```diff
@@ -644,6 +672,264 @@ def validate_condition(self):
...
+        os_disk = ''
+        prefix = 'ister-{}-'.format(str(config["Version"]))
+        target_dir = tempfile.mkdtemp(prefix=prefix)
+        disks = get_list_of_disks()
+        current = self._find_current_disk()
+        boot_part = ''
+        for disk in disks:
+            part_info = get_disk_info('/dev/{}'.format(disk))
+            for part in part_info['partitions']:
+                if part.get('type') and 'EFI' in part['type']:
```
bryteise: So similarly to making sure the current disk is not in the os_disk it should also not be in the boot disk. In fact I'd go so far as to say the boot_part and os_disk must be on the same disk for this feature because it is too much guessing otherwise. I'd also like comment to the block for the goal of finding a os disk and boot partition for that disk to repair as it is looking a little hard to follow at this point.

ister_test.py
```diff
@@ -4226,6 +4225,175 @@ def mock_sleep(sec):
...
+    chooseact = ister_gui.ChooseAction()
+    result = chooseact._find_current_disk()
+    subprocess.check_output = check_output_backup
+    if result:
+        raise Exception("Found root disk when there was none to be found")
+
+    commands_compare_helper(commands)
+
+
+def gui_mount_host_disk():
```
bryteise: This is a good test but misses bits that may be needs another test as if you didn't have a EFI partition on sdc it would have used sda instead.

ister_gui.py
```diff
@@ -644,6 +672,264 @@ def validate_condition(self):
...
+                    boot_part = part['name']
+                if part.get('type')                \
+                        and 'root' in part['type'] \
+                        and current not in part['name']:
+                    os_disk = part['name']
+                if os_disk and boot_part:
+                    break
+            if os_disk and boot_part:
+                break
+
```
bryteise: It may be good to break out the mount try to another function and make a test if os_disk has any data or not before calling that function and having a different error in that case.

Similarly for boot_part aside from it doesn't need to be an error just a notification that we couldn't find a boot partition to mount (vs an error mounting a boot partition we did find).

ister_gui.py
bryteise: This should go away at this point.

matthewrsj:
@bryteise changes made

matthewrsj:
I've deferred numbering the screen until we nail down a final screen order based on UX feedback.

ister_gui.py
```diff
@@ -644,6 +672,255 @@ def validate_condition(self):
...
+                                                 target])
+                    # only mount a linux root partition if it is Clear Linux OS
+                    if 'Linux root' in pattern:
+                        osf = '{}/usr/lib/os-release'.format(self.target_dir)
+                        with open(osf, 'r') as os_release:
+                            if 'clear-linux-os' in os_release.read():
+                                return part_found
+                            else:
+                                # maybe there is another root part on this disk
+                                continue
```
bryteise: Need to umount for all the mounted partitions that weren't a Clear Linux rootfs.

ister_test.py
```diff
@@ -4441,7 +4814,16 @@ def error(self, _):
...
+        gui_set_hw_time,
+        gui_find_current_disk_success,
+        gui_find_current_disk_failure,
+        gui_mount_host_disk_normal,
+        gui_mount_host_disk_no_boot_on_host,
+        gui_get_root_present,
+        gui_get_root_not_present,
+        gui_get_boot_present,
+        gui_get_boot_not_present,
+        gui_umount_host_disk
```
bryteise: Really nice testing.

matthewrsj:
@bryteise changes made

PR #30, opened 2016-09-08
Auto-detect and allow user to set http/https proxies

matthewrsj:
The gui displays both http and https proxies detected in the
environment and allows user to set them manually.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -400,10 +400,13 @@ def copy_os(args, template, target_dir):
         run_command("mount --bind {0}/var/tmp /var/lib/swupd"
                     .format(target_dir))
     swupd_env = os.environ
-    if template.get("Proxy"):
-        swupd_env["http_proxy"] = template["Proxy"]
-        swupd_env["https_proxy"] = template["Proxy"]
-        LOG.debug("proxy: {}".format(template["Proxy"]))
+    if template.get("HTTPSProxy"):
+        swupd_env["https_proxy"] = template["HTTPSProxy"]
+        LOG.debug("https_proxy: {}".format(template["HTTPSProxy"]))
+
+    if template.get("HTTPProxy"):
+        swupd_env["http_proxy"] = template["HTTPProxy"]
+        LOG.debug("http_proxy: {}".format(template["HTTPProxy"]))
 
     run_command(swupd_command, environ=swupd_env)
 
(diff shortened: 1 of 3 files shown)


ister.py
```diff
@@ -400,10 +400,13 @@ def copy_os(args, template, target_dir):
...
-        swupd_env["http_proxy"] = template["Proxy"]
-        swupd_env["https_proxy"] = template["Proxy"]
-        LOG.debug("proxy: {}".format(template["Proxy"]))
+    if template.get("HTTPSProxy"):
+        swupd_env["https_proxy"] = template["HTTPSProxy"]
+        LOG.debug("https_proxy: {}".format(template["HTTPSProxy"]))
+
+    if template.get("HTTPProxy"):
+        swupd_env["http_proxy"] = template["HTTPProxy"]
+        LOG.debug("http_proxy: {}".format(template["HTTPProxy"]))
```
bryteise: Hrm just realized that the template change should be added to the validate_template function (and tests added for that too).

ister_gui.py
```diff
@@ -713,7 +714,19 @@ def handler(self, config):
...
+        if os.environ.get('https_proxy'):
+            https_text = os.environ['https_proxy']
+        else:
+            https_text = ''
+
+        if os.environ.get('http_proxy'):
+            http_text = os.environ['http_proxy']
+        else:
+            http_text = ''
+
```
bryteise: Getting proxy information from the environment seems fine but I'm not sure that's really useful for an autorun installer where it shouldn't ever really be set. Is there some use case I'm missing?

ister_gui.py
```diff
@@ -854,8 +866,13 @@ def _reset_network(self, _):
 
     def _set_proxy(self, _):
         """Set the user defined proxy for the installer in the template"""
-        if self.installer_proxy.get_edit_text():
-            self.config['Proxy'] = self.installer_proxy.get_edit_text()
+        if self.https_proxy.get_edit_text():
+            self.config['HTTPSProxy'] = self.https_proxy.get_edit_text()
+            os.environ['https_proxy'] = self.https_proxy.get_edit_text()
```
bryteise: Similarly, I'm not sure that updating the environment here is doing anything.

ister_gui.py
matthewrsj: I set the environment variables so that subsequent commands (like curl below) can grab them from the environment. I would run into an issue trying to figure out which proxy to pass to curl if I allowed the user to set both.

ister_gui.py
matthewrsj: This is needed to detect previously set proxies when the page is refreshed.

matthewrsj:
Those tests have been added/updated

PR #31, opened 2016-09-15
Show swupd --install output in GUI

matthewrsj:
Users often complain about ister_gui not giving any status indication
when swupd starts up during the last step of the installation. This
patch displays the swupd output in the GUI alert so users can at
least see that the process is not frozen.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -60,7 +60,8 @@ LOG = None
 DEBUG = False
 
 
-def run_command(cmd, raise_exception=True, log_output=True, environ=None):
+def run_command(cmd, raise_exception=True, log_output=True, environ=None,
+                show_output=False):
     """Execute given command in a subprocess
 
     This function will raise an Exception if the command fails unless
@@ -78,9 +79,11 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None):
         for line in lines:
             decoded_line = line.decode('ascii', 'ignore').rstrip()
             output.append(decoded_line)
-            if log_output:
+            if show_output:
+                LOG.info(decoded_line)
+            elif log_output:
                 LOG.debug(decoded_line)
-        if proc.wait() != 0 and raise_exception:
+        if proc.poll() and raise_exception:
             decoded_line = proc.stderr.read().decode().rstrip()
             output.append(decoded_line)
             LOG.debug("Error {0}".format(decoded_line))
@@ -365,8 +368,10 @@ def add_bundles(template, target_dir):
     """
     bundles_dir = "/usr/share/clear/bundles/"
     os.makedirs(target_dir + bundles_dir)
-    for bundle in template["Bundles"]:
+    for index, bundle in enumerate(template["Bundles"]):
         open(target_dir + bundles_dir + bundle, "w").close()
+
+    LOG.info("Installing {} bundles (and dependencies)...".format(index + 1))
 
 
 def get_current_format():
@@ -408,7 +413,7 @@ def copy_os(args, template, target_dir):
         swupd_env["http_proxy"] = template["HTTPProxy"]
         LOG.debug("http_proxy: {}".format(template["HTTPProxy"]))
 
-    run_command(swupd_command, environ=swupd_env)
+    run_command(swupd_command, environ=swupd_env, show_output=True)
 
 
 class ChrootOpen(object):
--- a/ister_test.py
+++ b/ister_test.py
@@ -122,7 +122,7 @@ def run_command_wrapper(func):
     def wrapper():
         """run_command_wrapper"""
         def mock_run_command(cmd, _=None, raise_exception=True,
-                             log_output=True, environ=None):
+                             log_output=True, environ=None, show_output=False):
             """mock_run_command wrapper"""
             COMMAND_RESULTS.append(cmd)
             if not raise_exception:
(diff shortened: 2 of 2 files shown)


bryteise:
So I know we've had this before and folks were unhappy with the swupd output dump to their screen that was confusing.

I'm not sure anything has changed to go back to displaying this content.

matthewrsj:
I spoke with @rnesius and Auke regarding this and we agreed it is at least an *improvement* over displaying nothing at all. Users can at least see that something is happening and the process hasn't hung. As it is, it can take half an hour or more (considerably more if the network connection is bad) if they are installing all the optional bundles.

As far as confusing output, it isn't that bad. It tells you it is 
`Downloading pack <pack name> for version <version>` and then
`extracting pack`

This has to be better than nothing for half an hour.

bryteise:
That part isn't so much the problem as what comes after:

Calling post-update helper scripts.
[ERROR] cbm (src/bootman/files.c:L253): Invalid block device: /tmp/t
[ERROR] cbm (src/bootman/files.c:L137): Invalid block device: /tmp/t
[FATAL] cbm (src/bootman/sysconfig.c:L97): sysconfig insane: Missing root_uuid
[FATAL] cbm (src/bootman/update.c:L459): Failed to install bootloader
Possible filedescriptor leak : 4 (/usr/share/clear/update-ca/ClearLinuxRoot.pem)
Possible filedescriptor leak : 5 (socket:[14400903])
[ERROR] cbm (src/bootman/files.c:L253): Invalid block device: /tmp/t
[ERROR] cbm (src/bootman/files.c:L137): Invalid block device: /tmp/t
[FATAL] cbm (src/bootman/sysconfig.c:L97): sysconfig insane: Missing root_uuid
[FATAL] cbm (src/bootman/update.c:L459): Failed to install bootloader
Possible filedescriptor leak : 4 (/usr/share/clear/update-ca/ClearLinuxRoot.pem)
Possible filedescriptor leak : 5 (socket:[14400903])
[ERROR] cbm (src/bootman/files.c:L253): Invalid block device: /tmp/t
[ERROR] cbm (src/bootman/files.c:L137): Invalid block device: /tmp/t
[FATAL] cbm (src/bootman/sysconfig.c:L97): sysconfig insane: Missing root_uuid
[FATAL] cbm (src/bootman/update.c:L459): Failed to install bootloader
Possible filedescriptor leak : 4 (/usr/share/clear/update-ca/ClearLinuxRoot.pem)
Possible filedescriptor leak : 5 (socket:[14400903])
Fix successful

If we just want a progress bar I'd prefer to just track that a line was written to stdout and move a spinner or the like.

bryteise:
As an aside even when printing per line from swupd there is still an ~10 minute wait (on tempfs unsure how long it'd be on spinning rust) when all the downloads have completed and swupd is installing files where nothing is output from swupd.

matthewrsj:
I still think users would rather see output, even if the output contains errors (especially if the output contains errors?).

10 minute wait is better than 30. At least the user will see what is happening. They know they aren't waiting on network, etc.

ister.py
```diff
@@ -365,9 +368,11 @@ def add_bundles(template, target_dir):
     """
     bundles_dir = "/usr/share/clear/bundles/"
     os.makedirs(target_dir + bundles_dir)
-    for bundle in template["Bundles"]:
+    for index, bundle in enumerate(template["Bundles"]):
         open(target_dir + bundles_dir + bundle, "w").close()
 
+    LOG.info("Installing {} bundles (and dependencies)...".format(index))
```
bryteise: Should be index + 1

ister.py
```diff
@@ -408,7 +413,7 @@ def copy_os(args, template, target_dir):
         swupd_env["http_proxy"] = template["HTTPProxy"]
         LOG.debug("http_proxy: {}".format(template["HTTPProxy"]))
 
-    run_command(swupd_command, environ=swupd_env)
```
bryteise: This will cause any invocation (including nongui) to display all swupd output.

matthewrsj: Is this a problem? Wouldn't the output be helpful regardless? I can easily fix with an arg to ister that is only used by the gui, however.

matthewrsj:
@bryteise review ping

ister.py
```diff
@@ -78,9 +79,11 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None):
         for line in lines:
             decoded_line = line.decode('ascii', 'ignore').rstrip()
             output.append(decoded_line)
-            if log_output:
+            if show_output:
+                LOG.info(decoded_line)
+            elif log_output:
                 LOG.debug(decoded_line)
-        if proc.wait() != 0 and raise_exception:
```
bryteise: So this is interesting to me because I expected it to cause issues without the threading but when I looked closer at the gui code doing this I couldn't understand why. The reason this works (and why poll() vs wait() is actually a non issue here) is because the iterator is doing the work of reading through the pipe and will only terminate in this case once the pipe closes. Having this be poll is nicer than doing the comparison though I guess so I'll take it as is.

ister.py
bryteise: I'd rather avoid having to add an extra option that is essentially display swupd output... So I think this default is prefered.

PR #33, opened 2016-10-05
Network status text color indicates connectivity

matthewrsj:
Red text indicates there is no network connection to clearlinux.org
while green text indicates a successful connection. This indication
was identified as important by users during UX review.


bryteise:
Merged in a2f1317.

PR #34, opened 2016-10-05
Update the bundle select screen

matthewrsj:
- Update bundle select screen to show telemetrics

  If the user opted in, this screen shows the telemetrics bundle as part
  of the required bundle list. If the user did not opt in it does not
  appear. The user can also change their mind and this list is updated.

- Remove brackets from required bundle selections

  The square brackets surrounding the 'X' in front of the required
  bundles confused some users in the UX review. The required bundles
  looked like they were selectable because they were the same format
  as the optional bundles. Removing the brackets makes the required
  bundles look unselectable.


bryteise:
Merged in 3f3365f.

PR #37, opened 2016-10-07
Increase pycurl timeout from 1 second to 3 seconds

matthewrsj:
**Proposed** - on my home network, one second was not long enough

On slower networks one second is too short a time to receive a response
from clearlinux.org, causing a 'none detected' warning when there is
in fact a network connection. A three second timeout decreases the risk
of this false negative.


bryteise:
Merged in b789189.

PR #41, opened 2016-10-10
Redesign network screen

matthewrsj:
* Redesign network screen to group like config options 

  This was a confusing screen because it was not clear which buttons
  were related to which configuration field. They are not grouped in
  a logical manner and the < reset network > button only appears when
  it is relevant.

  Also remove the .8 second sleep on startup since this is no longer
  the first screen. The time the user spends on the splash screen will
  allow the network units to start up.

* Add optional DNS setting to static IP configuration


matthewrsj:
screenshot_2016-10-10_12-53-01

PR #42, opened 2016-10-12
Resize network configure options on message length

matthewrsj:
Dynamically set uwrid.Padding for network configuration buttons based
on the length of the message in the buttons. This creates buttons that
look like this:

```
< Button message >
```

rather than this:

```
< Button message                                                     >
```

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -170,6 +170,15 @@ def get_list_of_disks():
     return unmounted
 
 
+def ister_button(message, on_press=None, user_data=None,
+                 align='left', left=0, right=0):
+    width = len(message) + 4
+    button = urwid.Button(message, on_press=on_press, user_data=user_data)
+    button = urwid.AttrMap(button, 'button', focus_map='reversed')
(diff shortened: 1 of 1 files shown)


bryteise:
Was this only an issue with the proxy page? It would seem like this would be something all buttons would want to use.

matthewrsj:
So, there are other buttons, but they are on pages that are only for buttons. I thought keeping them the width of the page made sense. Another example is the nav bar, but I thought those buttons should all be the same size as well.

chooseact

bryteise:
Personally I'd rather keep the button look consistent across the UI but we probably want additional feedback from others before making a decision on that.

matthewrsj:
Looking through the screens, I'm agreeing with you that the buttons would look fine if they were consistently the length of the message. I'm working on this now.

matthewrsj:
@bryteise done. All buttons are now created to look consistent and use the same wrapper method to do so.

bryteise:
Merged in 2143879.

PR #44, opened 2016-10-12
Change ister GUI to a dark theme

matthewrsj:
This is a proposal to change ister_gui to a dark theme. Personally I like this better than light gray on dark blue. Please see screenshots below.

splash
telemetrics
netreq_nonefound
netreq_established
partalert
createuser
bundle_selector
(can you spot the bug ^^^?)
begininstall
installing

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -64,18 +64,18 @@ import ister
 
 PALETTE = [
     ('header', 'white', 'dark red', 'bold'),
-    ('banner', 'black', 'light gray'),
-    ('bg', 'black', 'dark blue'),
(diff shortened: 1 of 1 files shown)


bryteise:
Did we get that feedback from the UX/UI folks? This one is definitely harder on my eyes at least but I'm okay with the merge if this has some consensus around it.

matthewrsj:
Brent suggested a dark theme, which this is based off of. I like it but I'm not married to it. I'm just not a big fan of light gray on dark blue.

bryteise:
Good enough for me then.

PR #47, opened 2016-10-12
Remove X from required bundles

matthewrsj:
Per suggestion from @rnesius


bryteise:
Merged in d0c7743.

PR #52, opened 2016-10-24
Remove http_proxy settings since https_proxy is required

matthewrsj:
cURL requires a https_proxy (with http url scheme) to talk to
clearlinux.org. This removes all the http_proxy settings and
updates the example text to show a http url scheme.


bryteise:
Merged in f4a2727.

PR #49, opened 2016-10-14
Add ascii image of Clear Linux Penguin to SplashScreen

matthewrsj:
Remove introduction paragraph because this pushes much of the text off
the screen on widescreen monitors when the penguin is present.

Signed-off-by: Matthew Johnson <matthew.johnson@intel.com>

### qemu
penguin

### Bare metal
img_0039

PR #53, opened 2016-10-24
Center and pad the splash screen to make it look better

matthewrsj:

Since we are not merging #49 for 3.0, this at least makes the splash screen look nice.
plainsplash


bryteise:
Merged in d3c7e9a.

PR #54, opened 2016-10-26
Move static network validation to user edits, allow proxy unset

matthewrsj:
* Move static network validation to user edits

  As the user edits the static IP configuration boxes, their input is
  checked in real-time as they type. Once the field they are editing is
  detected as a valid interface/IP address, the field caption turns
  green, notifying the user of a valid configuration. The button to set
  the static configuration does not activate until all required fields
  are valid.

  Moving this logic to before button activation allows the removal of a
  lot of error handling complexity in the handler() function since the
  user is not allowed to press the button until the input is valid.
  Because of this, this patch also includes considerable simplification
  to the NetworkRequirements class.

* Allow user to unset https_proxy
  
  User can unset their previously entered https_proxy on the network
  requirements screen by (intuitively) deleting their entered proxy
  and pressing the set proxy button again.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1031,7 +1031,7 @@ class NetworkRequirements(ProcessStep):
         self.static_ready = False
         self.static_edits = {}
         self.set_static_edits()
-        self.reset= False
+        self.reset = False
 
     def set_static_edits(self):
         fmt = '{0:>20}'
@@ -1343,6 +1343,9 @@ class NetworkRequirements(ProcessStep):
         if self.https_proxy.get_edit_text():
             self.config['HTTPSProxy'] = self.https_proxy.get_edit_text()
             os.environ['https_proxy'] = self.https_proxy.get_edit_text()
+        else:
+            self.config.pop('HTTPSProxy', None)
+            os.environ.pop('https_proxy', None)
 
         raise urwid.ExitMainLoop()
 
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1023,24 +1023,20 @@ class NetworkRequirements(ProcessStep):
 
         # initiate necessary instance variables
         self.progress = urwid.Text('Step {} of {}'.format(cur_step, tot_steps))
-        self.interface_msg = 'Interface: '
-        self.ip_msg = 'IP address: '
         self.https_proxy = None
-        self.interface = None
-        self.error = None
         self.config = None
         self.ifaceaddrs = None
-        self.nettime = None
+        self.nettime = False
         self.static_set = False
-        self.static_check = False
         self.static_ready = False
         self.static_edits = {}
         self.set_static_edits()
+        self.reset= False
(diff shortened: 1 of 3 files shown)


matthewrsj:
### Screenshots

edits1
edits2
edits3
edits4

rnesius:
Very nice!  +1

ister_test.py
```diff
@@ -38,10 +38,19 @@
...
 # pylint: disable=W0702
 # Mock functions almost by definition do not make use of their inputs.
 # Fine if they don't use them
 # pylint: disable=W0613
+# We don't need docstrings for many of these methods, especially when they are
+# just mock methods anyways
+# pylint: disable=missing-docstring
+# Don't worry about protected access warnings since we want our unit tests to
+# test those methods
+# pylint: disable=protected-access
```
bryteise: pylint: disable=protected-access is listed twice.

matthewrsj: oops...

matthewrsj: fixed.

ister_gui.py
```diff
@@ -1049,20 +1045,59 @@ def set_static_edits(self):
...
+
+            if 'DNS' in key.caption and self.static_edits[key]:
+                try:
+                    ipaddress.ip_address(self.static_edits[key])
+                    self.dns_e.set_caption(
+                            ('success', '{0:>20}'.format('DNS (optional): ')))
+                except:
+                    self.dns_e.set_caption(
+                            '{0:>20}'.format('DNS (optional): '))
+                    reqd_found -= 1
```
bryteise: I am somewhat skeptical that the okay count is easier to understand than 4 different boolean variables but we can leave this as it is for now.

ister_gui.py
```diff
@@ -1086,45 +1121,19 @@ def handler(self, config):
...
-            Alert('Error!', self.error).do_alert()
-            self.error = ''
-            return self._action
+            self.static_button.original_widget.original_widget.set_label(
+                    ('ex', 'Set static IP configuration'))
+
+        # proxy settings may have been added during the self.run_ui() step,
+        # copy our config instance back to the config.
+        config = self.config
+        return self._action
```
matthewrsj: are you suggesting an error color?

rnesius: If not empty_string and not_valid, color red?

matthewrsj: I think the success color coupled with the button becoming activated is enough of an indication.

Another idea is to add an icon, like a green check mark, when the field is validated. This takes a little more work than just updating the caption color though, which is why I went with this option.

matthewrsj: @rnesius I'm not sure I'm a fan of immediate negative feedback when the user starts typing.

bryteise: So I'm going to merge this as is. We can go back later and change things if we get feedback.

PR #56, opened 2016-10-28
Add disk partition info to device selection step

matthewrsj:
Display disk partition information on device selection step so user is
able to verify they are installing to the correct disk before selecting
it. Previously, a user wasn't warned about disk partitions until the
alert before the installation started.

Signed-off-by: Matthew Johnson matthew.johnson@intel.com


matthewrsj:
### Screenshots

diskinfo
disknoinfo

bryteise:
Merged in b28d2354735ba3699f595d31cf35a4a0bd5289c4.

PR #61, opened 2016-12-27
Update Static IP Step for host

matthewrsj:
Use the same logic to validate IP addresses for the host static IP step
as is used in the installer static IP configuration. Separates the logic
out into a NetworkControl class for reuse.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -51,9 +51,9 @@ import threading
 import sys
 import pprint
 import time
-import ipaddress
 import tempfile
 import shutil
+import ipaddress
 import netifaces
 import pycurl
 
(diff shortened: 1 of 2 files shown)


ister_gui.py
```diff
@@ -1007,65 +998,145 @@ def _umount_host_disk(self, root_part, boot_part):
...
+        """Find active interface and ip address"""
+        # pylint: disable=E1103
+        addrs = {}
+        mask = ''
+        af_inet = netifaces.AF_INET
+        for if_name in netifaces.interfaces():
+            ifaddrs = netifaces.ifaddresses(if_name)
+            if af_inet in ifaddrs:
+                ip_addrs = ifaddrs[af_inet][0]
+                if ip_addrs['addr'] and '127.0.' not in ip_addrs['addr']:
```
bryteise: Why 127.0 and not just 127?

ister_gui.py
```diff
@@ -1007,65 +998,145 @@ def _umount_host_disk(self, root_part, boot_part):
...
+        for interface in addrs:
+            if interface.startswith('e'):
+                return (interface, addrs[interface], mask)
+
+    def find_dns(self):
+        """Find active DNS server by searching /etc/resolv.conf if it exists"""
+        content = ''
+        if os.path.exists('/etc/resolv.conf'):
+            with open('/etc/resolv.conf', 'r') as resolv:
+                content = resolv.read()
```
bryteise: Why not just use readlines() since you are splitting by '\n' below?

bryteise:
Looks pretty good. Just a few questions.

ister_gui.py
matthewrsj: That was just me trying to be more specific in case 127 occurred somewhere else in the ip, but this would be better as

```python
if ip_addrs['addr'] and not ip_addrs['addr'].startswith('127'):
```

ister_gui.py
matthewrsj: You are right...

matthewrsj:
questions addressed.

PR #69, opened 2017-01-06
 Poll systemctl status output when restarting network

matthewrsj:
Instead of guessing at a time to sleep, poll systemctl status output
until network services are up.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1223,19 +1223,40 @@ class NetworkControl(object):
         except FileNotFoundError:
             pass
 
-        try:
-            subprocess.call(['/usr/bin/systemctl', 'restart',
-                             'systemd-networkd', 'systemd-resolved'])
-        except:
-            raise urwid.ExitMainLoop()
+        self._restart_networkd_resolved()
 
-        # give the network a chance to start up again
-        time.sleep(.1)
+        # set necessary flags, then reset static edit text.
         self.static_ready = False
         self.reset = True
         self.set_static_edits()
         raise urwid.ExitMainLoop()
 
+    def _restart_networkd_resolved(self):
+        """Restart the network services then poll systemctl status output until
+        both are back up"""
+        restarted = False
+        try:
+            # restart systemd-networkd and systemd-resolved
+            subprocess.call(['/usr/bin/systemctl', 'restart',
+                             'systemd-networkd', 'systemd-resolved'])
+            # check network status with increasing sleep times on each failure.
+            for i in [.1, 1, 2, 4]:
+                out = subprocess.check_output(['/usr/bin/systemctl',
+                                               'status',
+                                               'systemd-networkd',
+                                               'systemd-resolved'])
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -1423,6 +1423,7 @@ def _static_configuration(self, _):
         try:
             subprocess.call(['/usr/bin/systemctl', 'restart',
                              'systemd-networkd', 'systemd-resolved'])
+            time.sleep(.1)
```
bryteise: So could we possibly poll (look up ip a output or something) instead of sleeping here? There is another place or two with sleeps that it would be nice to avoid if possible (especially where if we don't wait long enough, there is a problem).

matthewrsj: What about just a `systemctl status systemd-networkd systemd-resolved` to make sure both are up? That seems easier than parsing ip a output to me.

bryteise: That'd be a really nice feature.

matthewrsj:
@bryteise I now poll `systemctl status networkd.service resolved.service`

ister_gui.py
```diff
@@ -1223,19 +1223,31 @@ def _reset_network(self, _):
...
         self.set_static_edits()
         raise urwid.ExitMainLoop()
 
+    def _restart_networkd_resolved(self):
+        """Restart the network services then poll systemctl status output until
+        both are back up"""
+        try:
+            subprocess.call(['/usr/bin/systemctl', 'restart',
+                             'systemd-networkd', 'systemd-resolved'])
+            while True:
```
bryteise: May want to sleep 1 and do a max of say 10 of these before saying there is an error?

matthewrsj: ok.

matthewrsj: sleep 1 is an awfully long time to sleep though. I wasn't running into any issues with sleeping a tenth of a second.

bryteise: Well you could do a step off increment, I mostly just care about a total maximum time that this gives up after and says there was a problem.

matthewrsj: step off increment? This is what I was going to do.

```python
for i in range(10):
    out = subprocess.check_output(['/usr/bin/systemctl',
                                    'status',
                                    'systemd-networkd',
                                    'systemd-resolved'])
    if out.decode('utf-8').count('Active: active (running)') == 2:
        # decrement i in case this is successful on the last try
        # with i == 9
        i -= 1
        break

    time.sleep(.1)

    # if == 9 we hit the try limit
    if i == 9:
        raise Exception('Unable to restart network services')

except Exception as err:
    Alert('Error!', err).do_alert()
    raise urwid.ExitMainLoop()
```

bryteise: Ah I'd just do for i in [.1, 1, 2, 4] sleep(i)

matthewrsj: ah, nice

matthewrsj:
@bryteise this now polls with increasing sleep times.

PR #70, opened 2017-01-17
Make several fixes to mount point screen

matthewrsj:
- Fix bug where removing partition after setting mount point caused
  exception.
- Only display 'auto format on root' message when user does not select
  the format option on root mountpoint.
- Remove old duplicate mount point if user enters that mount point
  again.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1793,7 +1793,7 @@ class SelectDeviceStep(ProcessStep):
 class SelectMultiDeviceStep(SelectDeviceStep):
     """UI to display the available disks"""
     def handler(self, config):
-        config["Disks"] = self.disks[:]
+        config["Disks"] = get_list_of_disks()
         other_options = ['Previous', 'Next']
         self.build_ui_widgets()
         self._ui = SimpleForm(u'Choose a drive to partition using cgdisk tool',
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -1935,6 +1949,14 @@ def handler(self, config):
             return exc
         return self._action
 
+    def _erase_dup_mount(self, point, part):
+        for idx, chc in enumerate(self.choices):
+            if '{} '.format(point) in chc and '{} '.format(part) not in chc:
```
bryteise: Could we strip strings before adding them to the map? This seems not good.

bryteise: Oh wait is that a post formatted map? Is the original data source updated later and we can't (re)create the formatting without duplicates later on?

matthewrsj: this is done to avoid finding the `/` mount point inside the `/boot` mountpoint and to make sure we are matching the entire partition name instead of a portion of another.

bryteise: Wait shouldn't you just use == then?

bryteise: But again that's for the preformatted handling. It seems awkward to do this post format.

matthewrsj: because `chc` is a long string that I am searching `in`

bryteise: Alrighty, taking as is.

PR #72, opened 2017-01-18
Add simple VM spinup script for testing

matthewrsj:
Add spinup.sh, which allows a user to spin up ./installer.img with
./OVMF.fd bios file to quick-test installer. This is a very simple
script meant to be sent to testers along with the other required files
so they can test the installer with minimal effort.

Requires vncviewer and qemu as dependencies.

Usage:

```bash
./spinup.sh <port number>
./spinup.sh     # port number defaults to 2233
```

### DIFF
--- a/spinup.sh
+++ b/spinup.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -e
+
+x(){
+    echo -- "$@" >&2
+    "$@"
+}
+
+runinst(){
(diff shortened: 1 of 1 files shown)


spinup.sh
```diff
@@ -0,0 +1,34 @@
+x(){
```
bryteise: Should have #!/bin/bash to start. Also set -e is probably a good idea.

matthewrsj: fixed

PR #76, opened 2017-01-27
Require sysadmin-basic bundle when admin user created

matthewrsj:
When an administrated user is created, root login is disabled. For this
user to be able to do much of anything in the system, they need sudo.
sudo will not be included unless the sysadmin-basic bundle is selected.
This patch makes sysadmin-basic required only when an administrative
user is set, and makes it optional when a user is not set (login as
root).

Ister also remembers the user's last selection before the bundle was
made mandatory. This allows the user to change their mind about user
configuration without worrying about deselecting the sysadmin-basic
bundle again.

Resolves CLEAR-1843


ister_gui.py
```diff
@@ -2129,6 +2124,32 @@ def handler(self, config):
                 bundle for bundle in self.required_bundles
                 if not bundle.get('name') == 'telemetrics']
 
+        sysadmin_reqd = any(bundle['name'] == 'sysadmin-basic'
+                            for bundle in self.required_bundles)
+        sysadmin_basic = {'name': 'sysadmin-basic',
+                          'desc': 'Run common utilities useful for managing a '
+                                  'system'}
+        if config.get('Users') and not sysadmin_reqd:
```
bryteise: So any user means admin user?

ister_gui.py
```diff
@@ -2146,6 +2167,12 @@ def handler(self, config):
                         else:
                             if bundle in config['Bundles']:
                                 config['Bundles'].remove(bundle)
+
+        # update only if user had chance to unselect
+        if not any(bundle['name'] == 'sysadmin-basic'
+                   for bundle in self.required_bundles):
+            self.sysadmin_basic_state = 'sysadmin-basic' in config['Bundles']
```
bryteise: So config['Bundles'] holds the default_bundles. Now default_bundles may have had sysadmin-basic removed from it if there is a user created but sysadmin-basic would have been added to config['Bundles'] separately in that case (fun logic chain to follow). I may have missed it but I don't see a way for sysadmin-basic to get back into default_bundles if a user ended up changing selections nor do I understand how self.bundles gets reset back to a default state either.

ister_gui.py
matthewrsj: The options are `Create an administrative user` or `No user creation (login as root)`

ister_gui.py
matthewrsj: config['Bundles'] holds the bundles that will be installed on the system, so all user selections plus required bundles. default_bundles are just selected by default, not included by default.

I don't add sysadmin-basic back into default_bundles (I don't include it as selected by default again) because I want it to be removed from the required_bundles list and placed into the optional bundles list in the same state it was in before it was *made* to be required...

So I track the state the sysadmin-basic bundle was in before being made required and if I have to put it back, I assign it the same state it was before requiring it. I acknowledge that is hard to follow.

matthewrsj:
Reordered screens @bryteise

PR #77, opened 2017-01-31
Make use of swupd's --path option for repair

matthewrsj:
The 'Repair' option in Ister was failing to complete the repair due to a
failed network connection in the chroot. Instead of chrooting and trying
to set up a network, the path to the mounted host OS can be provided to
swupd via the --path option. This removes the complicated chroot
requirements and fixes the 'Repair' option.

AKA: RTFM

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -855,45 +855,11 @@ class ChooseAction(ProcessStep):
             self._umount_host_disk(root_part, boot_part)
             self._action = 'return'
             return
-        if os.path.exists('/run/systemd/resolve/resolv.conf'):
-            try:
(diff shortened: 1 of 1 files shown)


matthewrsj:
I seriously can't believe I missed this the first time.

bryteise:
Just confirming, is /boot already mounted under the target as well?

matthewrsj:
Well, not after this change, but that is how it used to be done for a repair.

matthewrsj:
Sorry, yes, /boot is mounted. still..

bryteise:
Oh good. thanks.

PR #79, opened 2017-02-16
RFE: Add sudo users to the sudo group, not to the sudoers config file

ikeydoherty:
Full explanation in the commit itself. It may be that you intend for this behaviour, but in case
it isn't, here's an alternative. ^^

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -534,19 +534,17 @@ def disable_root_login(target_dir):
 
 
 def setup_sudo(user, target_dir):
-    """Append user to sudoers file
+    """Add user to sudo (wheel) group
 
     This function will raise an Exception on finding an error.
     """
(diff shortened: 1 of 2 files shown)


ikeydoherty:
^ lol @ looking for rake

bryteise:
Yea, I need to take the time to setup travis here, I welcome pull requests for that =P.

bryteise:
Nice use of the chroot wrapper btw.

ikeydoherty:
Well tbh it took me a while to figure out how to make the test happy. At which point I became most impressed, that's a nice test system you have :D

bryteise:
When you think about the amount of pain mocking gives you lol. Thank goodness for python at least it is much simpler.

ikeydoherty:
Yeah, something like this in C would be a god send ..

PR #82, opened 2017-02-17
Add configuration files for Travis CI

matthewrsj:
Ignore this until it works.

update: this works

Several tests had to be modified to properly mock internal functions and subprocess calls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "3.5"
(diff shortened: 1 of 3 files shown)


matthewrsj:
@bryteise this works now, some tests were missing mocks and they were `just working` because of the system it was running on for local testing.

bryteise:
Thanks so much, also great finding of mock escape =).

PR #84, opened 2017-03-10
Rework statedir handling for physical installs

bryteise:
When doing physical installations, the statedir argument was being
largely ignored incorrectly due to the use of the bind mount. Rework the
bind mount usage and at the same time fixup the creation of the
directory to set the mode to only be accessable by the owner (root).
Also fixup cleanup to handle the potential new location of the statedir.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -38,22 +38,24 @@
 import argparse
 import ctypes
 import json
+import logging
+import netifaces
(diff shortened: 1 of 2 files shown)


bryteise:
@matthewrsj Mind giving this a lookover?

PR #85, opened 2017-03-27
Add timeout messages for network service/detection

matthewrsj:
Add timeout messages for network service startup and network detection
if network information is not found. These messages are removed/replaced
when network is set up.

Does not change GUI flow - no tests modified or added.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -258,10 +258,14 @@ def required_bundles(config):
 def network_service_ready():
     """Check network status with increasing sleep times on each failure"""
     for i in [.1, 1, 2, 4]:
-        out = subprocess.check_output(['/usr/bin/systemctl',
-                                       'status',
-                                       'systemd-networkd',
-                                       'systemd-resolved'])
+        try:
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -1110,6 +1114,11 @@ def update_widgets(self):
         if not dns_res:
             dns_res = 'none found'
 
+        if 'none found' in ''.join([interface_res, mask_res, ip_res, dns_res]):
```
bryteise: The text 'none found' seems to conflict with the set_text message. Is there a way to get them better aligned or maybe just explain away my confusion.

matthewrsj: set_text is is adding the refresh message to the header iff 'none found' was in a result, which is almost always because there was not enough time to set it.

The alternative is to wait until everything is found, but most people won't be needing to set static IP for the installer so why wait for that?

bryteise: Hrm alrighty, I guess the 'Detected network configuration' is what threw me when it keyed off of 'none found' as they seemed to be opposing messages.

Issue #87, opened 2017-04-21
NetworkRequirement's _network_connection() test should add test for dns


bryteise:
Also should test against cdn.download.clearlinux.org uri for both dns and curl.

PR #88, opened 2017-04-26
Allow configuration to dictate network testing

bryteise:
Instead of always testing the clearlinux website as the network
available standard, make use of the content and version urls passed in
as arguments. At the same time update the clearlinux urls to reflect the
cdn to the update content instead of the main page similar to how the
swupd-client itself works.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1421,10 +1421,12 @@ class NetworkControl(object):
 
 class NetworkRequirements(ProcessStep):
     """UI to verify and configure network connectivity to
-    https://www.clearlinux.org for the installer"""
+    the version and content urls for the installer"""
     # pylint: disable=R0902
-    def __init__(self, cur_step, tot_steps):
+    def __init__(self, cur_step, tot_steps, content_url, version_url):
         super(NetworkRequirements, self).__init__()
+        self.content_url = content_url
+        self.version_url = version_url
         left = 8  # padding desired left of buttons, modifiable
         # configure proxy section
         self.proxy_header = urwid.Text('Proxy Settings')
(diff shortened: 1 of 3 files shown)


bryteise:
While not exactly doing the prescribed method of fixing #87 this seems like the correct way to resolve the issue. @matthewrsj looking for your review =).

ister_gui.py
```diff
@@ -2733,7 +2742,9 @@ def __init__(self, *args, **kwargs):
     def _init_actions(self):
         # auto install assumed
         keyboard_select = KeyboardSelection()
-        network_requirements = NetworkRequirements(1, 6)
+        network_requirements = NetworkRequirements(1, 6,
+                                                   self.args['contenturl'],
+                                                   self.args['versionurl'])
```
matthewrsj: If you look on line 2722, self._init_actions() is called before self.args is set to kwargs on line 2733. Using self.args in _init_actions results in an `'Installation' object has no attribute 'args'` error.

bryteise: Ah indeed, which reminds me since I have you here, what is your gui test method for kicking the tires outside of the unit tests?

matthewrsj: I make an image using ister
```
python3 ister.py -t installer-config.json
```

I load up the local GUI and ister changes
```
./update_installer.sh installer.img
```

I run the instance with qemu
```
sudo qemu-system-x86_64 -enable-kvm -m 1024 -vnc 0.0.0.0:0 -cpu host -drive file=installer.img,if=virtio,aio=threads -net nic,model=virtio -net user,hostfwd=tcp::2233-:22 -smp 2 -bios <path to bios file>
```

Then I just use vncviewer.

I have that qemu command mapped to a bash function. If I make changes to the gui I just run `./update_installer.sh installer.img` again.

bryteise: Ah lovely, thanks!

PR #92, opened 2017-07-14
Persist installer network config to target screen

matthewrsj:
When the user enters a static IP configuration for the installer, the
target static IP configuration fields are pre-populated with the
settings. This prevents the user from having to enter the configuration
twice.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -2603,6 +2603,10 @@ class StaticIpStep(ProcessStep):
         self.gateway_s = ''
         self.dns_s = ''
         self.config = {}
+        self.ip_res = ''
+        self.mask_res = ''
+        self.dns_res = ''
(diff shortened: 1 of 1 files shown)


bryteise:
So the rational makes sense to me but I am not so sure about one part. Since the installer image is not stateless, the static ip will persist to other installation attempts which seems clumsy and we might want to consider how we handle being more stateless in the future. +1 for this change though.

matthewrsj:
So, the static IP setting has been persistent since we allowed users set it for the installer image on the network configuration screen.

I like the idea of making this more stateless in the future, should be easy enough to just run some cleanup on boot. That could either be done as a service file or in ister itself. In this case we just need to remove 10-en-static-network and everything is reset back to DHCP.

PR #93, opened 2017-07-17
Make pylint and pycodestyle fixes

matthewrsj:
Make Ister pylint and pycodestyle clean. Many exceptions are removed, and those that remain are renamed to use their title instead of error number.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -18,28 +18,17 @@
 # Boston, MA 02110-1301 USA
 
 # The array as default value is too handy here to worry about
-# pylint: disable=W0102
+# pylint: disable=dangerous-default-value
 # Intended manipulation of internal state of the UI
-# pylint: disable=W0212
+# pylint: disable=protected-access
 # global is handy here for now, but this could be fixed up
-# pylint: disable=W0603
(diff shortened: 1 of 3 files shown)


ister_gui.py
```diff
@@ -403,8 +420,8 @@ def construct(self):
                                  height=('relative', PERCENTAGE_H))
         self._ui = urwid.AttrMap(self._ui, 'banner')
 
-    # This callback is registered with urwid.MainLoop and gives us the
-    # opportnity to intercept keystrokes and handle things like tabs.
+    # This callback is registered with urwid.MainLoop and gies us the
+    # opportnity to intercept keystrokes and handle things lke tabs.
```
bryteise: Introduced two typos here.

bryteise:
Nice cleanup. +1 aside from the typos.

ister_gui.py
phmccarty: you could also s/opportnity/opportunity as well :)

matthewrsj:
@bryteise typos fixed

matthewrsj:
@bryteise everything is how it should be. ister_test.py cleaned up as well.

PR #94, opened 2017-07-18
Add interface configuration field for target

matthewrsj:
Add configuration field for the interface pattern in target static
network configuration screen. This allows users to specify which NIC the
static configuration is for on machines with multiple NICs. Without this
feature users were seeing the same static IP address being set for every
NIC on the server, which caused connection failures.


matthewrsj:
@phmccarty since @bryteise is out, do you have some time to look at this?

PR #99, opened 2017-08-29
NVME partition support

matthewrsj:
Instead of stripping all digits from the partition name and calling it
the device name, use the output of lsblk -no pkname <partition> to find
the real device name.

This fixes bugs that occur when the device name actually has digits in
it (e.g. /dev/nvme0n1).

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -107,6 +107,32 @@ def get_disk_info(disk):
     return info
 
 
+def get_part_devname(part):
+    """
+    Return the parent device of the partition
+
+    If called on a device name, we could get a list of identical strings for
+    each partition the device contains:
+
+    $ lsblk -no pkname /dev/sda
+            # note the blank line here
+    sda
+    sda
+    sda
+
+    Just return the first one
+    """
+    cmd = ['/usr/bin/lsblk', '-no', 'pkname', os.path.join('/dev', part)]
+    try:
+        # need to strip out leading blank line for the case that lsblk is
+        # called on a device name
+        output = subprocess.check_output(cmd).decode('utf-8').strip()
+    except subprocess.CalledProcessError:
+        return None
+
+    return output.splitlines()[0]
+
+
 def get_list_of_disks():
     """"
     Queries for the available disks discarding the installer root disk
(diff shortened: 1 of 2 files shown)


ister_gui.py
```diff
@@ -350,8 +376,20 @@ def search_swap(choices, config, mount_d):
                 output = output[idx + len(pttr):]
                 output = output[:output.index('"')]
                 if output == 'swap':
-                    disk = ''.join(x for x in part if not x.isdigit())
+                    # first try to set using lsblk -no pkname part
+                    disk = os.path.join('/dev', get_part_devname(part))
+                    # if that fails, fallback to old (unreliable) method
```
bryteise: Has failure been something that has happened? I'm inclined to just say we have a bug in the lsblk toolchain path if we ever see it fail rather than potentially introduce bad configuration data.

ister_gui.py
```diff
@@ -350,8 +376,20 @@ def search_swap(choices, config, mount_d):
...
+                    disk = disk or ''.join(x for x in part if not x.isdigit())
                     part = part[len(disk):]
+                    # must be an int
+                    try:
+                        # strip out prefix or other indicators
+                        part = int(''.join(x for x in part if x.isdigit()))
+                    except ValueError:
+                        Alert('Partition error',
+                              'Unable to detect partition number for {}'
+                              .format(part)).do_alert()
```
bryteise: Would it be reasonable to print out the partition type and disk as well do you think? Could assist in error reports.

ister_gui.py
matthewrsj: No, failure hasn't happened, but the old method works most of the time.

How about reporting an Alert() when the first one fails that the install may not succeed?

bryteise: I suspect if the first one fails the second one is going to have the same/worse problems (wrong path and something is funky with hardware for instance). But if this seems more likely to succeed leave it in.

matthewrsj: My thought here was to include the fallback in the case you mentioned where we have a bug in the lsblk toolchain. In that case, the lsblk command might fail but we can still detect partition info using the old method.

bryteise: Alrighty.

ister_gui.py
```diff
@@ -2253,7 +2310,7 @@ def _save_config(self, config, mount_d):
             if not mount_d[point]['format']:
                 if 'TYPE="' in subprocess.check_output(
                         'blkid | grep {0}'
-                        .format(disk+part), shell=True).decode('utf-8'):
+                        .format(disk+prefix+part), shell=True).decode('utf-8'):
```
matthewrsj: @bryteise this is the change that was needed. Prefix needs to hang around for this call, but partition needs to be a number or a string representation of a number by the time ister.py takes over.

bryteise: Ah okay, great.

PR #100, opened 2017-08-31
Add screen to append to kernel cmdline

matthewrsj:
Allow the user to append parameters to the kernel cmdline. This screen
shows an example of the sane defaults used in Clear Linux for native
kernel. The appended parameters are added to <target>/etc/kernel/cmdline
and clr-boot-manager update is run to append the parameters to the boot
entry.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -44,6 +44,7 @@ import tempfile
 import ipaddress
 import signal
 import time
+import itertools
 import netifaces
 import pycurl
 import urwid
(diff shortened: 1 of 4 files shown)


matthewrsj:
Eh, a pylint fix for the nvme test made it into this commit. I can take it out and add it in another commit if you wish.

bryteise:
No need, can't really tell ;).

matthewrsj:
Forgot to update expect tests again. Hold on.

ister_gui.py
```diff
@@ -2880,33 +2930,48 @@ def __init__(self, *args, **kwargs):
...
+        config_hostname = ConfigureHostname(next(current), total)
+        # User configuration meta-step
+        step = next(current)
+        confirm_user = ConfirmUserMenu(step, total)
+        user_configuration = UserConfigurationStep(step, total)
+        bundle_selector = BundleSelectorStep(next(current), total)
+        # Networking meta-step
+        step = next(current)
+        confirm_dhcp = ConfirmDHCPMenu(step, total)
+        static_ip_config = StaticIpStep(step, total)
```
bryteise: I like the use of the counter but it gets a little confusing to me when the code alternates between next(current) and step when setting up the orderings. Maybe some reordering or more descriptive variable names as I see this being a possible issue for somebody who wants to change the list order in the future figuring out where to insert what potentially.

matthewrsj: I have to save next(current) to a variable if I want to use the int more than once before it is incremented.

I'll change the variable names and make the comments more verbose.

bryteise: Thanks

matthewrsj:
@bryteise expect tests updated

PR #101, opened 2017-09-06
Read available disks every time screen is reached

matthewrsj:
Instead of reading the list of disks only once when ister_gui is
initializing, read them every time the disk partition screen is reached.
This fixes a bug where slow-loading devices have not loaded by the time
the available disks are determined and are therefore never discovered by
ister, causing the installation to be impossible.

The choice was made to read the disks every time instead of just until
some devices are found since some devices may load faster than others.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -2068,10 +2068,12 @@ class SelectDeviceStep(ProcessStep):
         super(SelectDeviceStep, self).__init__()
         self._clicked = None
         self._actions = None
-        self.disks = get_list_of_disks()
+        self.disks = []
         self.progress = 'Step {} of {}'.format(cur_step, tot_steps)
 
     def handler(self, config):
+        # look every time, slow-loading or plug-in devices may show up late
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -2068,10 +2068,12 @@ def __init__(self, cur_step, tot_steps):
         super(SelectDeviceStep, self).__init__()
         self._clicked = None
         self._actions = None
-        self.disks = get_list_of_disks()
+        self.disks = []
         self.progress = 'Step {} of {}'.format(cur_step, tot_steps)
 
     def handler(self, config):
+        # just look every time, slow-loading devices may show up late
+        self.disks = get_list_of_disks()
```
bryteise: Seems fine, I just wonder if maybe we should have a button to refresh the device list instead/too?

matthewrsj: That would be a good addition considering it would a user could later plug in a USB device and want to refresh the screen.

matthewrsj:
Refresh button added.

Issue #103, opened 2017-09-26
provide a way to run raw shell commands post install.


steverweber:
`validate_postnonchroot_template` is limiting what can run to only script files that can be found on the filesystem. this is painful when trying to performing some custom commands after a pxe boot install. I'm 
 forced to change the initrd to hack in some script.. would be much cleaner and direct if the shell commands could be included in the `ister.json`.

it would be helpful if `validate_postnonchroot_template` did no checks at all or another option like  `PostNonChrootRaw` was added.

running raw commands would be nice.
```
"PostNonChroot": ["bash -c 'mkdir -p $0/usr/local/maintenance; cd $0/usr/local/maintenance; curl https://xxxxxx.com/linux/postinstall.tar.gz | tar -xz'"]
```

steverweber:
my attempt failed because the command args are split on space
```
subprocess.Popen(cmd.split(" "),
```

perhaps more friendly would be giving the user a shell and let bash split the args.
```
something like.
PostNonChrootRaw:
    for cmd in PostNonChrootRaw:
        subprocess.Popen(['/bin/bash', '-c' , cmd]
```

bryteise:
So do you want a non-chroot and chroot version of this post install shell command runner? 

Also giving the shell commands the chroot location in an environment variable seems like the best way to go in the non-chroot case, is that fitting with your use case?

steverweber:
I think the chroot version is more helpfull but having the none chroot version would be a nice extra.

agree passing in the chroot path as an env var is the way to go.

If you do go to add these. Perhaps also add a PreBashCmds.  This would allow some way to shim in support for odd things like perhaps raid drivers.

bryteise:
I have code working for the both chroot and non-chroot post handlers. I'm curious when exactly you'd like the pre handlers to be run though. I'd guess you'd get the most flexability from having them run directly after template validation but that could be too early. From my perspective having commands run after validation and before any actual processing would be enough to support raid. Any more details for your use case would be handy here.

steverweber:
I didn't have a true use case for the pre command.

debian has something like this `d-i preseed/early_command` it runes before the partitioner, so after validation should be a good place.

bryteise:
@steverweber #105 created in response to this issue.

PR #105, opened 2017-09-27
Pre post handlers

bryteise:
Add additional script and shell command handlers to:
* Run shell commands prior to installation starting
* Run post install shell commands both in and out of the installation chroot
* Run post install scripts inside the installation chroot (running scripts outside of the chroot was already supported)

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -646,7 +646,7 @@ def post_install_nonchroot(template, target_dir):
 
     if not template.get("PostNonChroot"):
         return
-    LOG.info("Running post scripts")
+    LOG.info("Running post non-chroot scripts")
     for script in template["PostNonChroot"]:
         run_command(script + " {}".format(target_dir))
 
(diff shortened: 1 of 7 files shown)


bryteise:
This addresses #103.

ister.py
```diff
@@ -73,7 +73,11 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None,
     try:
         LOG.debug("Running command {0}".format(cmd))
         sys.stdout.flush()
-        proc = subprocess.Popen(cmd.split(" "),
+        if shell:
+            full_cmd = ["sh", "-c", cmd]
```
matthewrsj: is this any different than just adding `shell=True` to the `Popen` call?

ister.py
```diff
@@ -640,11 +654,49 @@ def post_install_nonchroot(template, target_dir):
...
+    """Run chroot post install scripts
+
+    All post scripts must be executable.
+    """
+    if not template.get("PostChroot"):
+        return
+    LOG.info("Running post scripts")
+    with ChrootOpen(target_dir) as _:
+        for script in template["PostChroot"]:
+            run_command(script + " {}".format(target_dir))
```
matthewrsj: Is it documented that the PostChroot script must accept a single target_dir argument?

matthewrsj: I guess I would have the same question for the PostNonChroot script - I've never made use of these myself.

pre-post.json
```diff
@@ -0,0 +1,16 @@
...
+    "FilesystemTypes" : [ { "disk" : "test.img", "partition" : 1, "type" : "vfat" },
+                          { "disk" : "test.img", "partition" : 2, "type" : "ext4" } ],
+    "PartitionMountPoints" : [ { "disk" : "test.img", "partition" : 1, "mount" : "/boot" },
+			       { "disk" : "test.img", "partition" : 2, "mount" : "/" } ],
+    "Version": "latest",
+    "Bundles": ["os-core"],
+    "PreInstallShell": ["echo preinstall shell command"],
+    "PostNonChroot": ["./post-non-chroot.sh"],
+    "PostNonChrootShell": ["touch $ISTER_CHROOT/post-non-chroot-shell"],
+    "PostChroot": ["/post-chroot.sh"],
```
matthewrsj: should there be a leading dot here?
```
"PostChroot": ["./post-chroot.sh"],
```
Or not since this is in the chroot?

ister.py
steverweber: agree would be more clean
```
subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=environ, shell=True)
```

steverweber: personal i would drop the extra shell argument and just use a shell by default... the installer worked fine when i did that.

ister.py
bryteise: Ah PostChroot shouldn't get the target_dir passed in that was an oversight. And sadly there is not much documentation at all on the script requirements. Making the README useful is quite a work item that needs to be looked at.

ister.py
matthewrsj: I agree with @bryteise that shell would not be good to use everywhere, but I think it makes sense for these scripts and commands.

pre-post.json
bryteise: Oh wait sorry I misread your quote for the current code. There shouldn't be a dot because it is the chroot.

matthewrsj: Right OK, +1 here then.

PR #106, opened 2017-10-02
Use dd not qemu img

bryteise:
Switch to dd instead of qemu-img for creating virtual disks. Also enable stderr output to be output to the console even if the program didn't exit with a non-zero return code.


matthewrsj:
Yay magic numbers... Yay fewer dependencies! +1

PR #107, opened 2017-10-02
Stop staticly salting users passwords

bryteise:
A static salt is doing nothing here (too small, publicly available),
use the default salt instead.


matthewrsj:
+1

phmccarty:
+1

matthewrsj:
*statically

bryteise:
Pushed with spelling fix.

PR #110, opened 2017-10-03
Fix sparse file creation

bryteise:
When using dd to create a sparse file seek should be set to be the
offset and count should be set to 0.


matthewrsj:
+1, this seems to follow docs I've found for creating sparse files.

PR #111, opened 2017-10-11
Resolve dd and swap bugs

bryteise:
Fixes:
* gui swap detection
* dd not using sparse file creation
* dd to use megabyte offsets for parted

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -147,7 +147,7 @@ def create_virtual_disk(template):
     # partition sizes specified in MiB (maybe eventually switch to doing
     # math based on each disk or virtual disks sector size).
     image_size += 40
-    command = "dd if=/dev/zero of={0} bs=1024 count={1} seek={1}".\
+    command = "dd if=/dev/zero of={0} bs=1024 count=0 seek={1}".\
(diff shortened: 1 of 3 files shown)


ister.py
```diff
@@ -142,12 +142,10 @@ def create_virtual_disk(template):
...
-    # Increase buffer by 10x to avoid insufficient space due to dd using 1K
-    # sector sizes and parted is using the native sector sizes and getting
-    # partition sizes specified in MiB (maybe eventually switch to doing
-    # math based on each disk or virtual disks sector size).
-    image_size += 40
-    command = "dd if=/dev/zero of={0} bs=1024 count={1} seek={1}".\
+    # Increase buffer by 1MB to give parted wiggle room due to dd using 1K
+    # sector sizes and parted is getting partition sizes specified in MiB.
+    image_size += 1024
+    command = "dd if=/dev/zero of={0} bs=1024 count=0 seek={1}".\
```
matthewrsj: same as #110 ?

bryteise: Ugh messed up my branches it seems. Well I'll just take this one since it is literally the same commit id =P.

PR #112, opened 2017-11-10
Set default uid to 1001 for consistency with gnome panel

matthewrsj:
The gnome settings UI adds new users with the 1001 uid. Ister should be
consistent with the default desktop environment installed via the
desktops bundle.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -482,12 +482,9 @@ def create_account(user, target_dir):
     """
 
     os.makedirs(target_dir + "/home", exist_ok=True)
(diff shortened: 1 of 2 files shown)


bryteise:
Honestly, I'd prefer that we configured our gnome settings UI to use user 1000 since this is just what the raw useradd command is doing. Trying to mimic ister behavior to an arbitrary tool seems like the wrong way of doing this.

matthewrsj:
I misunderstood the problem we were discussing. Closing.

PR #115, opened 2017-11-17
Use target swap space during install if available

bryteise:
In order to allow installation in more memory constrained situations,
make use of the install target's swap partition if available before
running swupd.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -289,6 +289,9 @@ def create_filesystems(template):
                         .format(base_dev, fst["partition"]))
         if "disable_format" not in fst:
             run_command(command)
+            if fst["type"] == "swap":
+                run_command("swapon {0}{1}".format(dev, fst["partition"]),
+                            raise_exception=False)
 
 
 def setup_mounts(template):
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -289,6 +289,8 @@ def create_filesystems(template):
                         .format(base_dev, fst["partition"]))
         if "disable_format" not in fst:
             run_command(command)
+            if fst["type"] == "swap":
+                run_command("swapon {0}{1}".format(dev, fst["partition"]))
```
matthewrsj: Maybe don't want to fail on this if it doesn't work for some reason. You could pass `raise_exception=False`

bryteise: Seems reasonable though that seems like it could imply disk issues.

matthewrsj: reading the `swapon` man page there are several reasons that `swapon` could fail, not all of them indicating that the disk is bad. I don't see how we would hit any of these cases right now, but maybe in the future, like if we support btrfs again (`swapon may not work correctly when using a swap file with some versions of btrfs`).

bryteise: Well but that's a swap file on btrfs not a swap partition (this portion isn't using swap files at all). But I'll push the change.

matthewrsj:
+1

PR #117, opened 2017-11-20
Add "may require Refresh" message to static IP settings

matthewrsj:
Add a message warning users that setting the static IP on the installer
may require a "Refresh" of the screen to see the changes reflected in
the UI. This approach is better than blocking the reload of the screen
until the new setting have taken place.


bryteise:
+1

Issue #116, opened 2017-11-18
ister must ignore mmcblk0rpmb as device


fenrus75:
On some systems a /dev/mmcblk0rpmb exists. fdisk -l fails on this device, crashing ister and aborting the install.

the rpmb device is some sort of secure partition not for OS use, and should just be ignored, not crashed on

bryteise:
Fixed via #118

PR #118, opened 2017-11-21
Catch exceptions from check_output calls

bryteise:
A subprocess.check_output call is causing a crash in ister making it
impossible to complete the install process. To fix this and avoid
potential future issues, add exception handling to all check_output
calls in the gui.


bryteise:
Fixes #116

matthewrsj:
Ah, good change +1

Issue #119, opened 2017-12-19
Error installing to BXT_GP board


matthewrsj:
Originally reported by @zyuan2x in https://github.com/clearcontainers/runtime/issues/869

The following error is encountered when installing Clear Linux to a BXT_GP board:

`fdisk: cannot open /dev/mmcblk0rpmb: Input/output error`

@bryteise I thought you resolved this issue in https://github.com/bryteise/ister/pull/118/commits/92305ebd92dab070e7c6dafc0acd56b6cafe7547

zyuan2x:
Does "Add exception handling to get_disk_info #120" patch in the latest image clear-19780-installer.img.xz  ?

matthewrsj:
There hasn't been a new release of ister yet. I'll do that today and the new version should be in this afternoon's build. I'm going to leave this issue open until we can be sure the issue is resolved.

zyuan2x:
Install Clear Linux on BXT_GP:
Result: Installed successfully with the image clear-19810-installer.img.xz

zyuan2x:
@matthewrsj  You can free to close this issue. Very thanks for your support.

PR #120, opened 2017-12-19
Add exception handling to get_disk_info

bryteise:
Similar to 92305ebd92dab070e7c6dafc0acd56b6cafe7547, except instead of
handling the error in the callers of get_disk_info, fix it in the
function directly since all but one of the callers were checking for
empty results anyway.

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -81,7 +81,10 @@ def get_disk_info(disk):
     """Return dictionary with disk information"""
     info = {'partitions': []}
     cmd = ['/usr/bin/fdisk', '-l', disk]
-    output = subprocess.check_output(cmd).decode('utf-8')
+    try:
+        output = subprocess.check_output(cmd).decode('utf-8')
+    except:
+        return info
     lines = output.split('\n')
     expr = re.compile('^Device')
     # discard header...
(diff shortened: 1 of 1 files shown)


bryteise:
@matthewrsj I believe this is the right fix for #119. Do you have time to make a release for this?

ister_gui.py
```diff
@@ -81,7 +81,10 @@ def get_disk_info(disk):
     """Return dictionary with disk information"""
     info = {'partitions': []}
     cmd = ['/usr/bin/fdisk', '-l', disk]
-    output = subprocess.check_output(cmd).decode('utf-8')
+    try:
+        output = subprocess.check_output(cmd).decode('utf-8')
+    except:
+        pass
```
matthewrsj: Shouldn't we return early here instead of trying to parse the output?

bryteise: I was going to do that but decided that the code is already in theory handling empty output and I don't like returning early if I don't need to just for readability.

matthewrsj: What's wrong with explicitly returning early? We know we want to return the initialized empty `info` dict when this errors out, but a `pass` means we are expecting the rest of the function to be a no-op on the dict. I think returning `info` here can help prevent bugs if we modify the rest of the function at some point in the future.

bryteise: Alrighty, I'll update the patch.

matthewrsj:
+1

PR #123, opened 2018-01-03
Add option for moving statedir into targetdir

gtkramer:
The current install approach favors disk space by allowing for the state
dir to exist outside of the target dir.  However, the two directories
can exist across block device boundaries.  To favor time to install,
adding an option to relocate the state dir into the target dir so they
are located within the same block device.  This can reduce install time
by as much as 60%.  Once the install has completed, the state dir is
removed from the target dir.

This has a couple of side effects.  When this option is used, the
template needs to allocate space for the state dir in addition to actual
content.  Additionally, this option cannot be used in conjunction with
the statedir option, as the state dir in this case is automatically
derived.  Because of these side effects, this option is not enabled by
default.

Signed-off-by: George T Kramer <george.t.kramer@intel.com>

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -1435,7 +1435,7 @@ def main():
         install_os(args)
     except Exception as exep:
         LOG.debug("Failed: {}".format(repr(exep)))
-        # todo: Add arg for loglevel to -v
+        # TODO: Add arg for loglevel to -v
         # And change this to trigger on DEBUG level
         if DEBUG:
             traceback.print_exc()
--- a/ister.py
+++ b/ister.py
@@ -423,6 +423,8 @@ def copy_os(args, template, target_dir):
     swupd_command += " --contenturl={0}".format(args.contenturl)
     swupd_command += " --versionurl={0}".format(args.versionurl)
     swupd_command += " --format={0}".format(args.format)
+    if args.fast_install:
+        args.statedir = "{0}/tmp/swupd".format(target_dir)
     swupd_command += " --statedir={0}".format(args.statedir)
     if args.cert_file:
         swupd_command += " -C {0}".format(args.cert_file)
@@ -440,6 +442,9 @@ def copy_os(args, template, target_dir):
         LOG.debug("https_proxy: {}".format(template["HTTPSProxy"]))
 
     run_command(swupd_command, environ=swupd_env, show_output=True)
+
+    if args.fast_install:
+        run_command("rm -rf {0}".format(args.statedir))
 
 
 class ChrootOpen(object):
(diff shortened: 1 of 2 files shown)


gtkramer:
I'm not sure if I filled out the list of commands correctly for the added test, copy_os_fast_install_good.  Do I have enough booleans listed?

ister.py
```diff
@@ -103,7 +103,7 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None,
                 LOG.debug(decoded_line)
 
         if proc.poll() and raise_exception:
-            if len(stderr) > 0:
```
matthewrsj: style fix-up/changes should go in a separate commit from the one that changes or adds functionality.

ister.py
```diff
@@ -412,7 +412,7 @@ def get_current_format():
 def copy_os(args, template, target_dir):
     """Wrapper for running install command
     """
-    LOG.info("Starting swupd. May take several minutes")
```
matthewrsj: Reason for this? Regardless it should not be in the same commit as a functional change.

ister.py
```diff
@@ -423,6 +423,8 @@ def copy_os(args, template, target_dir):
     swupd_command += " --contenturl={0}".format(args.contenturl)
     swupd_command += " --versionurl={0}".format(args.versionurl)
     swupd_command += " --format={0}".format(args.format)
+    if args.fast_install:
+        args.statedir = "{0}/swupd".format(target_dir)
```
matthewrsj: This will give you `<target_dir>/swupd` instead of `<target_dir>/var/lib/swupd` which is what I assume you want.

ister.py
```diff
@@ -1435,7 +1443,7 @@ def main():
         install_os(args)
     except Exception as exep:
         LOG.debug("Failed: {}".format(repr(exep)))
-        # todo: Add arg for loglevel to -v
+        # TODO: Add arg for loglevel to -v
```
matthewrsj: Comment update should be in a different commit as well.

ister.py
matthewrsj: Ah, you remove it later on after install is complete, so it doesn't really matter.

ister.py
```diff
@@ -1418,6 +1423,9 @@ def handle_options():
     parser.add_argument("-S", "--statedir", action="store",
                         default="/var/lib/swupd",
                         help="Path to swupd state dir")
+    parser.add_argument("-F", "--fast-install", action="store_true",
```
bryteise: To avoid confusing users, please put this in a [mutual-exclusion](https://docs.python.org/3/library/argparse.html#mutual-exclusion) with the statedir argument.

ister.py
bryteise: And do this for all cases of TODO needing to be corrected please!

ister.py
bryteise: But in case we decide there is going to be a swupd directory in / someday I'd probably use tempfile.mkdtemp with a prefix using the date or something to ensure this won't cause problems.

bryteise: Actually, to simply life in testing, just include the date in the directory name.

ister.py
gtkramer: I grepped through the code base in the repo and this was the only case.

ister.py
gtkramer: I forgot to do this.  I'll update the PR.

gtkramer:
Rebasing before picking up the work again.  Not ready to merge.

gtkramer:
@bryteise @matthewrsj the latest round of changes are ready and have been tested.  This cuts the installation time down for os-clr-on-clr by 55% from 22 minutes to 10 minutes.

gtkramer:
Actually, let's hold off on this.  Let's chat in person before thinking about what's next for this PR.

PR #124, opened 2018-01-16
Move openssh-server to the default list

bryteise:
Also update installer config image size


matthewrsj:
Also, I note that this shouldn't require a change to our expect scripts since, although the bundle will move in the bundle list, the bundle list will be the same length, requiring the same amount of tabs to navigate through.

bryteise:
Yea, I checked the validate release script before pushing just to be sure.

PR #125, opened 2018-01-25
ister: Add encryption support to ister

josuedhg:
ister.py: add options for encryption using ister.json
ister_gui.py: add options for encryption in the user interface
and add bootloader-extras bundle when root encryption is selected
ister_test.py: test json options for ister.py

Signed-off-by: Josue David Hernandez <josue.d.hernandez.gutierrez@intel.com>

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -58,6 +58,7 @@ from urllib.error import URLError, HTTPError
 from urllib.parse import urlparse
 from contextlib import closing
 import netifaces
+import pycryptsetup
 
 LOG = None
 
@@ -288,6 +289,22 @@ def create_filesystems(template):
 0657fd6d-a4ab-43c4-84e5-0933c84b4f4f"
                         .format(base_dev, fst["partition"]))
         if "disable_format" not in fst:
+            if "encryption" in fst:
+                c_dev="{0}{1}".format(dev, fst["partition"])
+                c = pycryptsetup.CryptSetup(device=c_dev)
+                c.luksFormat(cipher = "aes",
+                    cipherMode= "xts-plain64", keysize = 512, hashMode = "sha256")
+                c.addKeyByPassphrase(fst["encryption"]["passphrase"],
+                    fst["encryption"]["passphrase"])
+                c.activate(name=fst["encryption"]["name"],
+                    passphrase=fst["encryption"]["passphrase"])
+                if fst.get("options"):
+                    command = "{0} {1} /dev/mapper/{2}".format(fs_util[fst["type"]],
+                                              fst["options"],
+                                              fst["encryption"]["name"])
+                else:
+                    command = "{0} /dev/mapper/{1}".format(fs_util[fst["type"]],
+                                              fst["encryption"]["name"])
             run_command(command)
             if fst["type"] == "swap":
                 run_command("swapon {0}{1}".format(dev, fst["partition"]),
@@ -371,7 +388,13 @@ c12a7328-f81f-11d2-ba4b-00a0c93ec93b"
                         .format(base_dev, part["partition"]))
         if part["mount"] != "/":
             run_command("mkdir -p {0}{1}".format(target_dir, part["mount"]))
-        run_command("mount {0}{1} {2}{3}".format(dev,
+        if "encryption" in part:
+            run_command("mount /dev/mapper/{0} {1}{2}".format(
+                                                 part["encryption"]["name"],
+                                                 target_dir,
+                                                 part["mount"]))
+        else:
+            run_command("mount {0}{1} {2}{3}".format(dev,
                                                  part["partition"],
                                                  target_dir,
                                                  part["mount"]))
@@ -749,6 +772,11 @@ def cleanup(args, template, target_dir, raise_exception=True):
     if template.get("dev"):
         run_command("losetup --detach {0}".format(template["dev"]),
                     raise_exception=raise_exception)
+    for dev_entry in template['PartitionMountPoints']:
+        if 'encryption' in dev_entry:
+            c_dev = "{0}{1}".format(dev_entry['disk'], dev_entry['partition'])
+            c = pycryptsetup.CryptSetup(name=dev_entry['encryption']['name'])
+            c.deactivate()
 
 
 def get_template_location(path):
@@ -1326,7 +1354,7 @@ def parse_config(args):
     return config
 
 
-def install_os(args):
+def install_os(args, template):
     """Install the OS
 
     Start out parsing the configuration file for URI of the template.
@@ -1336,8 +1364,6 @@ def install_os(args):
     This function will raise an Exception on finding an error.
     """
     target_dir = None
-    configuration = parse_config(args)
-    template = get_template(configuration["template"])
 
     validate_template(template)
     try:
@@ -1371,10 +1397,8 @@ def install_os(args):
         cleanup(args, template, target_dir, False)
 
 
-def handle_logging(level, logfile):
+def handle_logging(level, logfile, shandler=logging.StreamHandler(sys.stdout)):
     """Setup log levels and direct logs to a file"""
-    shandler = logging.StreamHandler(sys.stdout)
-
     # Apparently the LOG object's level trumps level of handler?
     LOG.setLevel(logging.DEBUG)
 
@@ -1395,10 +1419,10 @@ def handle_logging(level, logfile):
         LOG.addHandler(fhandler)
 
 
-def handle_options():
+def handle_options(sys_args):
     """Setup option parsing
     """
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(prog='ister')
     parser.add_argument("-c", "--config-file", action="store",
                         default=None,
                         help="Path to configuration file to use")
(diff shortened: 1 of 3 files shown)


ister_gui.py
```diff
@@ -2395,10 +2474,18 @@ def _save_config(self, config, mount_d):
                     output = ''
                 if 'TYPE="' in output:
                     config['FilesystemTypes'][-1]['disable_format'] = True
+            if "encryption" in mount_d[point]:
+                config['FilesystemTypes'][-1]["encryption"] = {
+                    "passphrase": mount_d[point]['encryption'],
```
bryteise: Since this is writing the passphrase to disk (and may be reused or given to somebody else) we should make sure to securely delete this file after the installer has finished running.

josuedhg: the whole template I suppose

ister_gui.py
```diff
@@ -2395,10 +2474,18 @@ def _save_config(self, config, mount_d):
...
+                    "name" : "encrypted_dev" }
             config['PartitionMountPoints'].append({
                 'disk': disk,
                 'partition': part,
                 'mount': point})
+            if "encryption" in mount_d[point]:
+                config['PartitionMountPoints'][-1]["encryption"] = {
```
bryteise: So thinking about this a bit, I'd like to know how you would want to handle multiple partitions and disks being used. Do you expect all of them on a system to share the same encryption key or should they be able to have different keys per partition? I'm actually not sure if there is a use case that makes much sense to me to support from the installer for different keys so I was thinking about having Encryption be an optional toplevel item in the template with a passphrase and a list of dictionaries with disk + partition in them. Thoughts?

ister_gui.py
```diff
@@ -2395,10 +2474,18 @@ def _save_config(self, config, mount_d):
...
+                    "name" : "encrypted_dev" }
+                config["Bundles"].append("bootloader-extras")
```
matthewrsj: This needs to be called out in the bundle screen if we are adding an additional bundle. We can't just silently add bundles, even if it is a relatively small bundle (about 3.6 MB).

matthewrsj:
This will also probably need an update to maninstall.expect.

ister_gui.py
josuedhg: I was thinking in the screen when you decide if a partition will be formatted, just add there the option for encrypt and if the type of partition is /root apply this process if not add information to the crypttab but this could be a future improvement, at this time I am just handling root partitioning, crypttab support hash for passphrase then every partition could have a passphrase

ister_gui.py
josuedhg: I will do

matthewrsj: It should be added to the required bundles list if this option is set, and the optional bundles list if not - or maybe absent entirely. Additionally there needs to be some messaging on **this** screen to tell the user encryption requires the bootloader-extras bundle.

josuedhg: of course I already seen that :)

ister_gui.py
josuedhg: I am working in change how ister.py is called in order to keep passphrase hide, I mean there is an import for ister in the top of ister_gui.py then call install_os must be to easy, I will change some thinks in order to make this posible, tha way how ister.py and ister_gui.py are called will be keeped

josuedhg:
@matthewrsj maninstall use automatic partitions, this encryption option are only available in manual partitioning at the moment, I plan to add it to automatic partition and auto install, but this patch is not doing that

ister_gui.py
josuedhg: done

ister_gui.py
josuedhg: done

ister.py
```diff
@@ -1434,21 +1460,23 @@ def handle_options():
...
+    args = parser.parse_args(sys_args)
     return args
 
 
 def main():
     """Start the installer
     """
     global LOG
-    args = handle_options()
+    args = handle_options(sys.argv[1:])
```
bryteise: So if we are not passing sys.argv[0] then we should probably set prog= in the ArgumentParser initialization (prog="ister" would make sense to me).

ister.py
```diff
@@ -1434,21 +1460,23 @@ def handle_options():
...
 
     LOG = logging.getLogger(__name__)
     handle_logging(args.loglevel, args.logfile)
 
     try:
-        install_os(args)
+        configuration = parse_config(args)
+        template = get_template(configuration["template"])
```
bryteise: Since we get the template here it would make sense to me to call validate_template right after instead of inside the install_os function.

josuedhg: if we use template inside install_os we will force the creation of the template in the hard disk when ister is used by iste_gui, by this way we only provide a template if we run this in the command line.

ister.py
josuedhg: updated

ister.py
bryteise: Okay, yea I'd say from an ister as API perspective install_os is where we expect folks to call into and we don't expect them to validate the template they create as we won't want to just run off with that. Agreed.

ister_gui.py
```diff
@@ -910,7 +995,7 @@ def __init__(self):
     def get_next_step(self, action):
         """Returns the process step matched by action"""
         if action == 'Exit':
-            exit(0)
+            sys.exit(0)
```
bryteise: These changes to exit being called confuse me a little. Mind explaining this one?

ister_gui.py
```diff
@@ -3158,32 +3264,37 @@ def automatic_install(self):
...
+        alertLogger.setLevel(logging.INFO)
+        ister.LOG.addHandler(alertLogger)
+
+        open(ister_log, 'w').close()
+        fhandler = logging.FileHandler(ister_log)
+        fhandler.setLevel(logging.DEBUG)
+        formatter = logging.Formatter(
+            '%(asctime)s-%(levelname)s: %(message)s')
+        fhandler.setFormatter(formatter)
+        ister.LOG.addHandler(fhandler)
```
bryteise: I feel like this should be something ister handles as it is quite a lot of goo that shouldn't be exposed to the caller. I'll add an issue to the github for this though and I can come back to it later.

ister_gui.py
josuedhg: exit(0) is a helper for the python interactive session and sys.exit(0) is what must be used for scripts

ister_test.py
```diff
@@ -131,6 +139,50 @@ def good_template_string_partitions():
...
+                    COMMAND_RESULTS.append(keysize)
+                if hashMode:
+                    COMMAND_RESULTS.append(hashMode)
+
+            def addKeyByPassphrase(self, passphrase1, passphrase2):
+                COMMAND_RESULTS.append(
+                    "{0} - {1}".format(passphrase1, passphrase2))
+
+            def activate(self, name='', passphrase=''):
+                pass
```
bryteise: What do you think about adding to the COMMAND_RESULTS something like "activating", "deactivating instead of just pass here?

ister_gui.py
josuedhg: yes I did by this way in order to not change a lot of the ister.py but it could be handled by ister.py just requesting for a handler and using a default if it not exist

ister_test.py
josuedhg: ammm, ok, it is a good idea

ister_test.py
```diff
@@ -987,6 +1088,58 @@ def mock_listdir(directory):
...
+                                          "partition": 2}],
+                "FilesystemTypes": [{"disk": "sda", "partition": 1,
+                                     "type": "vfat",
+                                          "encryption" : {
+                                            "name" : "mapper_name",
+                                            "passphrase" : "123"}},
+                                    {"disk": "sda", "partition": 2,
+                                     "type": "ext4"}],
+                "Version": 10}
+    commands = ["ister-10-",
```
bryteise: Hrm I'm not sure where this 'ister-10-' is coming from, mind explaining this one for me?

josuedhg: me neither I took it from the other setup_mounts_good

josuedhg: I just add additional output expected with the changes that I provide the template

ister_gui.py
josuedhg: I will change this in order to provide handle it in a better way

ister_test.py
bryteise: Okay, I looked it up, just something setup_mounts is doing for the tempdir (ister-$version-)

bryteise:
Okay this looks good to me. @matthewrsj Could you give this another pass over on the gui side as I remember you mentioning there were some tricky details in part of that.

bryteise:
@matthewrsj ping

matthewrsj:
/me looks around for his test setup...

matthewrsj:
So, how did you test this? This causes the installer image to need `pycryptsetup`

matthewrsj:
@josuedhg need to add `pycryptsetup` to `requirements.txt`

josuedhg:
@matthewrsj pycryptsetup cannot be installed with pip

matthewrsj:
Ok, tested this and everything looks good from a GUI experience. +1

matthewrsj:
@josuedhg you need to connect with @mdhorn if this is a necessary feature going forward as I believe he is writing a new version of the installer.

mdhorn:
The plan is to ensure we have compliance with the current installer for the rewrite, but it would be helpful if you want to [create an issue in the new repository](https://github.intel.com/iclr/clr-installer/issues). This is only accessible inside of Intel at the moment until we get approved for the open sourcing the first release.

bryteise:
Resolved merge conflict and merged.

PR #126, opened 2018-01-25
Disable swap devices enabled by ister

bryteise:
Turn off swap devices that were enabled by ister during image
creation/installation. This resolves an issue where loop devices could
not be cleaned up after image creation.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -739,6 +739,13 @@ def cleanup(args, template, target_dir, raise_exception=True):
         run_command("rm -fr {}".format(target_dir),
                     raise_exception=raise_exception)
 
+    # Turn off any swap devices we enabled
+    for fst in template["FilesystemTypes"]:
+        (dev, _) = get_device_name(template, fst["disk"])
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -739,6 +739,13 @@ def cleanup(args, template, target_dir, raise_exception=True):
         run_command("rm -fr {}".format(target_dir),
                     raise_exception=raise_exception)
 
+    # Turn off any swap devices we enabled
+    for fst in template["FilesystemTypes"]:
+        (dev, _) = get_device_name(template, fst["disk"])
+        if fst["type"] == "swap":
+            run_command("swapoff {0}{1}".format(dev, fst["partition"]),
```
matthewrsj: can you just do swapoff --all to turn them all off?

bryteise: You can but if it is the image creation machine you disabled their swap too =P.

matthewrsj: Oh, of course

PR #127, opened 2018-02-15
Skip Network Screen when active

mdhorn:
If the network connection is active, skip the Networking
screen completely.

Signed-off-by: Mark D Horn <mark.d.horn@intel.com>

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -1608,7 +1608,7 @@ class NetworkRequirements(ProcessStep):
         if self.timeout:
             wired_req = urwid.Text(('warn', self.timeout))
         else:
-            if self._network_connection():
+            if self.network_connection():
                 wired_req = urwid.Text(['* Connection to the update server: ',
                                         ('success', 'established')])
             else:
@@ -1640,7 +1640,7 @@ class NetworkRequirements(ProcessStep):
     def run_ui(self):
         return self._ui.do_form(pop_ups=True)
 
-    def _network_connection(self):
+    def network_connection(self):
         """Check if connection to content and version urls are available"""
         # pylint: disable=E1103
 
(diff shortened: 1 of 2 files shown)


bryteise:
So I like this for the automatic setup but I wonder if there might be a situation where the dynamic IP getting used might not be what the user needs in order to complete the install. I suppose in that case the user can switch to the console as a work around as that use case seems not obvious to me.

Thoughts? @matthewrsj too.

ister_gui.py
```diff
@@ -1608,7 +1608,7 @@ def build_ui_widgets(self):
         if self.timeout:
             wired_req = urwid.Text(('warn', self.timeout))
         else:
-            if self._network_connection():
+            if self.network_connection():
```
bryteise: Why the function rename?

mdhorn: @bryteise I renamed it because I need to access that member function on the object; assumed that we were following the leading underscore on private member functions.

bryteise: Ugh sorry, I missed the step.network_connection X(.

matthewrsj:
> So I like this for the automatic setup but I wonder if there might be a situation where the dynamic IP getting used might not be what the user needs in order to complete the install.

This is actually a big concern for some of our users (the IT group that supports Clear Linux). The dynamic IP setup was added here specifically for them and upon their request because they didn't want to/weren't able to switch TTYs for an installation (I don't know if this was due to running an expect script or if it was just a hassle). This group was sometimes doing a large number of installs and all of them needed to have a different static IP assigned. Setting it for the installer also pre-populates the fields for the target installation network configuration, which was also very useful for these users and specifically requested by them.

Now, I'm not entirely sure their network works on boot in their environment, but this is still a concern for a more general case.

What about making static IP setup a different screen, or an optional branch to a screen? I know there is a push to minimize the number of functional screens in the installer but the static IP settings are important. We should at least discuss how that should be addressed.

seanvk:
I was also wondering about the Static IP.
Weren't we going to do anything about the flow with auto-proxy detection and avoid the manual IP port edit?  Not trying to add to the list, because we have longer term changes to be considered.

mdhorn:
@matthewrsj Based on this information about the need for the Static IP, current leaning is to not implement this PR. Confirming with Sean and Rob.

seanvk:
@mdhorn in light of our chat and feedback from @matthewrsj and @bryteise on the actual IT group requirements, I think it best we abandon this pull request and shift gears to the long term changes.

PR #129, opened 2018-03-06
Add support for installing content with dnf

gtkramer:
In test scenarios, work to create a swupd update stream is not needed
and increases image generation time.  If content for a test scenario can
be supplied as a package repository, then dnf can be used to do an
install of packages in the same way mixer generates chroots.

To support a define-and-run approach with ister, template files may
define an additional "PackageManager" key, which may be either "dnf" or
"swupd", if not defined to preserve backwards compatbility.  In the case
where the package manager is swupd, the list of bundles remains the same
as it is today.  In the case where the package manager is dnf, the
bundle list is treated as a package list that is used during a dnf
install.

Much like how content and version URLs may be passed at the command line
for swupd, to pass package repository URLs, which there may be multiple,
a DNF configuration file must be passed instead with the new
--dnf-config option.

To mimic how swupd works after installing content, update triggers are
run if they exist.  Because using DNF makes ister distro-agnostic,
opting to make the image bootable by using a post non-chroot script as
opposed to calling clr-boot-manager update directly.

Signed-off-by: George T Kramer <george.t.kramer@intel.com>

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -403,23 +403,21 @@ def add_bundles(template, target_dir):
 def copy_os(args, template, target_dir):
     """Wrapper for running install command
     """
-    LOG.info("Starting swupd. May take several minutes")
+    package_manager = template["SoftwareManager"]
+    LOG.info("Starting {0}. May take several minutes".format(package_manager))
+    if package_manager == "swupd":
+        copy_os_swupd(args, template, target_dir)
+    elif package_manager == "dnf":
+        copy_os_dnf(args, template, target_dir)
+
+
+def copy_os_swupd(args, template, target_dir):
+    """Wrapper for running install command with swupd
+    """
     add_bundles(template, target_dir)
-    swupd_command = "swupd verify --install --path={0} " \
-                    "--manifest={1}".format(target_dir, template["Version"])
-    if shutil.which("stdbuf"):
-        swupd_command = "stdbuf -o 0 {0}".format(swupd_command)
-    if args.contenturl:
-        swupd_command += " --contenturl={0}".format(args.contenturl)
-    if args.versionurl:
-        swupd_command += " --versionurl={0}".format(args.versionurl)
-    if args.format:
-        swupd_command += " --format={0}".format(args.format)
+
     if args.fast_install:
         args.statedir = "{0}/tmp/swupd".format(target_dir)
-    swupd_command += " --statedir={0}".format(args.statedir)
-    if args.cert_file:
-        swupd_command += " -C {0}".format(args.cert_file)
 
     if template["DestinationType"] == "physical":
         os.makedirs(args.statedir, exist_ok=True)
@@ -428,15 +426,50 @@ def copy_os(args, template, target_dir):
         os.chmod("{0}/var/tmp".format(target_dir), stat.S_IRWXU)
         run_command("mount --bind {0}/var/tmp {1}"
                     .format(target_dir, args.statedir))
-    swupd_env = os.environ
-    if template.get("HTTPSProxy"):
-        swupd_env["https_proxy"] = template["HTTPSProxy"]
-        LOG.debug("https_proxy: {}".format(template["HTTPSProxy"]))
 
-    run_command(swupd_command, environ=swupd_env, show_output=True)
+    cmd = "swupd verify --install"
+    cmd += " --path={0}".format(target_dir)
+    cmd += " --manifest={0}".format(template["Version"])
+    if args.contenturl:
+        cmd += " --contenturl={0}".format(args.contenturl)
+    if args.versionurl:
+        cmd += " --versionurl={0}".format(args.versionurl)
+    if args.format:
+        cmd += " --format={0}".format(args.format)
+    cmd += " --statedir={0}".format(args.statedir)
+    if args.cert_file:
+        cmd += " --certpath={0}".format(args.cert_file)
+    if shutil.which("stdbuf"):
+        cmd = "stdbuf -o 0 {0}".format(cmd)
+    cmd_env = get_cmd_env(template)
+    run_command(cmd, environ=cmd_env, show_output=True)
 
     if args.fast_install:
         run_command("rm -rf {0}".format(args.statedir))
+
+
+def copy_os_dnf(args, template, target_dir):
+    """Wrapper for running install command with dnf
+    """
+    cmd = "dnf install --assumeyes"
+    if args.dnf_config:
+        cmd += " --config {0}".format(args.dnf_config)
+    cmd += " --installroot {0}".format(target_dir)
+    cmd += " {0}".format(" ".join(template["Bundles"]))
+    if shutil.which("stdbuf"):
+        cmd = "stdbuf -o 0 {0}".format(cmd)
+    cmd_env = get_cmd_env(template)
+    run_command(cmd, environ=cmd_env, show_output=True)
+
+
+def get_cmd_env(template):
+    """Get the environment variables with which commands will execute
+    """
+    cmd_env = os.environ
+    if template.get("HTTPSProxy"):
+        cmd_env["https_proxy"] = template["HTTPSProxy"]
+        LOG.debug("https_proxy: {}".format(template["HTTPSProxy"]))
+    return cmd_env
 
 
 class ChrootOpen(object):
@@ -760,7 +793,11 @@ def get_template(template_location):
     """Fetch JSON template file for installer
     """
     json_file = request.urlopen(template_location)
-    return json.loads(json_file.read().decode("utf-8"))
+    parsed_json = json.loads(json_file.read().decode("utf-8"))
+    # Supply default SoftwareManager value if not defined for backwards compatibility
+    if not parsed_json.get("SoftwareManager"):
+        parsed_json["SoftwareManager"] = "swupd"
+    return parsed_json
 
 
 def validate_layout(template):
@@ -934,6 +971,24 @@ def validate_disk_template(template):
     validate_partition_mounts(template, partition_fstypes)
 
 
+def validate_version_template(template):
+    """Attempt to verify the version is sane
+    """
+    version = template["Version"]
+    if isinstance(version, int) and version <= 0:
+        raise Exception("Invalid version number")
+    if isinstance(version, str) and version != "latest":
+        raise Exception("Invalid version string (must be 'latest')")
+
+
+def validate_softmgr_template(template):
+    """Attempt to verify the package manager is sane
+    """
+    package_manager = template["SoftwareManager"]
+    if package_manager != "dnf" and package_manager != "swupd":
+        raise Exception("Invalid package manager.  Use either swupd or dnf")
+
+
 def validate_user_template(users):
     """Attempt to verify all user related information is sane
 
@@ -1084,11 +1139,8 @@ def validate_template(template):
         raise Exception("Missing Bundles field")
     validate_type_template(template)
     validate_disk_template(template)
-    version = template["Version"]
-    if isinstance(version, int) and version <= 0:
-        raise Exception("Invalid version number")
-    if isinstance(version, str) and version != "latest":
-        raise Exception("Invalid version string (must be 'latest')")
+    validate_version_template(template)
+    validate_softmgr_template(template)
     if template.get("Users"):
         validate_user_template(template["Users"])
     if template.get("Hostname") is not None:
@@ -1426,6 +1478,9 @@ def handle_options():
                         help="Path to swupd state dir")
     group.add_argument("-F", "--fast-install", action="store_true",
                         help="Move swupd state dir inside image for a faster install")
+    parser.add_argument("-d", "--dnf-config", action="store",
+                        default=None,
+                        help="DNF configuration file for installing packages")
     args = parser.parse_args()
     return args
 
--- a/ister.py
+++ b/ister.py
@@ -400,29 +400,21 @@ def add_bundles(template, target_dir):
     LOG.info("Installing {} bundles (and dependencies)...".format(index + 1))
 
 
-def get_current_format():
-    """Find the format id (if any) on the current system
-    """
-    frmt = ""
-    with open("/usr/share/defaults/swupd/format", "r") as format_file:
-        frmt = format_file.read().strip()
-    return frmt
-
-
 def copy_os(args, template, target_dir):
     """Wrapper for running install command
     """
     LOG.info("Starting swupd. May take several minutes")
-    if not args.format:
-        args.format = get_current_format()
     add_bundles(template, target_dir)
     swupd_command = "swupd verify --install --path={0} " \
                     "--manifest={1}".format(target_dir, template["Version"])
     if shutil.which("stdbuf"):
         swupd_command = "stdbuf -o 0 {0}".format(swupd_command)
-    swupd_command += " --contenturl={0}".format(args.contenturl)
-    swupd_command += " --versionurl={0}".format(args.versionurl)
-    swupd_command += " --format={0}".format(args.format)
+    if args.contenturl:
+        swupd_command += " --contenturl={0}".format(args.contenturl)
+    if args.versionurl:
+        swupd_command += " --versionurl={0}".format(args.versionurl)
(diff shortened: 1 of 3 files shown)


gtkramer:
Unit tests still need updated.  Looking to get some early feedback on the initial approach.

dnf-boot.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+"$1"/usr/bin/clr-boot-manager update -p "$1"
```
bryteise: Inclined to call the file dnf-post-cbm.sh

ister.py
```diff
@@ -408,43 +411,62 @@ def get_current_format():
...
+        cmd = "swupd verify --install"
+        cmd += " --path={0}".format(target_dir)
+        cmd += " --manifest={0}".format(template["Version"])
+        cmd += " --contenturl={0}".format(args.contenturl)
+        cmd += " --versionurl={0}".format(args.versionurl)
+        cmd += " --format={0}".format(args.format)
+        cmd += " --statedir={0}".format(args.statedir)
+        if args.cert_file:
+            cmd += " -C {0}".format(args.cert_file)
+    else:
```
bryteise: I'm inclined to match specificly and error.

ister.py
```diff
@@ -408,43 +411,62 @@ def get_current_format():
         frmt = format_file.read().strip()
     return frmt
 
+def add_default_dnf_config(args):
```
bryteise: I'm inclined to not have this built into ister and just let dnf use /etc/dnf.conf if it exists by default or else error early in the process around validate template.

bryteise: Or if the dnf-conf passed in doesn't exist still error.

ister.py
```diff
@@ -408,43 +411,62 @@ def get_current_format():
...
-    if not args.format:
-        args.format = get_current_format()
-    add_bundles(template, target_dir)
-    swupd_command = "swupd verify --install --path={0} " \
-                    "--manifest={1}".format(target_dir, template["Version"])
+    package_manager = template["PackageManager"]
+    LOG.info("Starting {0}. May take several minutes".format(package_manager))
+
+    if package_manager == SWUPD:
+        if not args.format:
```
bryteise: Probably would reorganize this section in a different commit.

ister.py
```diff
@@ -408,43 +411,62 @@ def get_current_format():
...
+
+    if package_manager == SWUPD and template["DestinationType"] == "physical":
         os.makedirs(args.statedir, exist_ok=True)
         os.chmod(args.statedir, stat.S_IRWXU)
         os.makedirs("{0}/var/tmp".format(target_dir))
         os.chmod("{0}/var/tmp".format(target_dir), stat.S_IRWXU)
-        run_command("mount --bind {0}/var/tmp {1}"
-                    .format(target_dir, args.statedir))
-    swupd_env = os.environ
+        run_command("mount --bind {0}/var/tmp {1}".format(target_dir, args.statedir))
```
bryteise: Why reformat this line?

ister.py
```diff
@@ -408,43 +411,62 @@ def get_current_format():
...
+    
+    run_command(cmd, environ=cmd_env, show_output=True)
 
-    run_command(swupd_command, environ=swupd_env, show_output=True)
-
-    if args.fast_install:
+    if package_manager == SWUPD and args.fast_install:
         run_command("rm -rf {0}".format(args.statedir))
+    if package_manager == DNF and os.path.exists("{0}/usr/bin/update-helper".format(target_dir)):
+        run_command("{0}/usr/bin/update-helper {1}".format(target_dir, target_dir))
```
bryteise: This should be part of the post-dnf script.

ister.py
```diff
@@ -768,7 +790,12 @@ def get_template(template_location):
     """Fetch JSON template file for installer
     """
     json_file = request.urlopen(template_location)
-    return json.loads(json_file.read().decode("utf-8"))
+    parsed_json = json.loads(json_file.read().decode("utf-8"))
+    try:
```
bryteise: I think better to just do ->
if not parsed_json.get("PackageManager"):
    parsed_json["PackageManager"] = SWUPD

bryteise: And I'm curious why this is done here instead of in validate_template?

dnf-boot.sh
gtkramer: Maybe pkg-post-update.sh?  Since it would also be running the update helper and cbm update?  pkg since in theory, we could add support for additional package managers?

ister.py
gtkramer: Match specifically?  Maybe we could default the cert file to the one on the system for swupd?

ister.py
gtkramer: Having this built-in would allow for a similar default for dnf analogous to what swupd has with its content and version URLs.  It's kinda ugly though, but maybe less ugly than the alternative of passing repo URLs and then generating a conf in /tmp somewhere?

ister.py
gtkramer: No need at all :)

ister.py
gtkramer: Ah, yes.  I'll move this in.

ister.py
gtkramer: I was thinking of defaulting the PackageManager to swupd as close to the source as possible so that validate_template would be strictly checking if the template that was supplied was all right instead of possibly modifying it with a default value.

gtkramer: Ah, so that's what the get function is used for.  I was initially doing `if not parsed_json["PackageManager"]` but getting a KeyError exception.  I Googled badly for that one.

ister.py
gtkramer: Maybe I could break this out into two different functions?  Something like
```
def install_os(args):
    if package_manager == SWUPD:
        install_os_swupd(args)
    elif package_manager == DNF:
        install_os_dnf(args)
```

dnf-boot.sh
bryteise: Ah yes sure

ister.py
bryteise: Ah sorry I mean if SWUPD elif DNF else raise error

ister.py
bryteise: So the built-ins for swupd urls are essentially legacy.
It would be nice to remove them and None the argument by default. Then we can only add those arguments to swupd when they are provided instead of every time.

ister.py
bryteise: That'd be fine but I meant the reorganization of the swupd options being appended to the cmd sorry.

ister.py
bryteise: Okay, that's fine with me then.

ister.py
gtkramer: Ah, yessir!  I'll add this in.

ister.py
gtkramer: Ok.  I'll do a separate commit for this as pre-work for this commit.

ister.py
gtkramer: I'll create a few different commits to organize things a little more nicely as a result of supporting multiple package managers.

gtkramer:
I've made some additional commits that hopefully help to tidy up the code base.  I'll take a look at the failing unit tests tomorrow.

gtkramer:
Something else to maybe consider?  I noticed that as a result of doing a dnf install of the packages listed in bundle definitions alone, the resulting image is not quite equivalent to an image that a swupd verify --install creates.  Namely, the /usr/share/clear folder is missing.  Besides running `update-helper` and `clr-boot-manager update` after a dnf install to get a bootable image, is there anything else that *should* be done?

bryteise:
It will not be equivalent you are correct. If you want the bundles files in /usr/share/clear you should poke them in the post script but they aren't important for the dnf installation, just for using swupd.

gtkramer:
Before names are cemented in stone for all time, I was wondering if PackageManager is the best term for both dnf and swupd?  dnf is a package manager, but swupd is a bundle manager, right?  Maybe SoftwareManager would be more accurate?

bryteise:
SoftwareManager is fine

gtkramer:
All tests have been updated.  I don't think there's anything else on my end.  How do things look?

ister.py
```diff
@@ -413,8 +413,6 @@ def copy_os(args, template, target_dir):
     """Wrapper for running install command
     """
     LOG.info("Starting swupd. May take several minutes")
-    if not args.format:
```
bryteise: Similarly to content and version urls, I think we can just remove the grabbing the format from the filesystem entirely.

ister.py
```diff
@@ -414,20 +414,11 @@ def copy_os(args, template, target_dir):
...
-                    "--manifest={1}".format(target_dir, template["Version"])
-    if shutil.which("stdbuf"):
-        swupd_command = "stdbuf -o 0 {0}".format(swupd_command)
-    if args.contenturl:
-        swupd_command += " --contenturl={0}".format(args.contenturl)
-    if args.versionurl:
-        swupd_command += " --versionurl={0}".format(args.versionurl)
-    swupd_command += " --format={0}".format(args.format)
+
+    add_bundles(template, target_dir)
```
bryteise: This is happening twice now? Seems like a mistake.

ister.py
```diff
@@ -431,12 +439,12 @@ def copy_os(args, template, target_dir):
     cmd = "swupd verify --install"
     cmd += " --path={0}".format(target_dir)
     cmd += " --manifest={0}".format(template["Version"])
-    cmd += " --format={0}".format(args.format)
-    cmd += " --statedir={0}".format(args.statedir)
     if args.contenturl:
         cmd += " --contenturl={0}".format(args.contenturl)
     if args.versionurl:
         cmd += " --versionurl={0}".format(args.versionurl)
+    cmd += " --format={0}".format(args.format)
```
bryteise: Seems odd to move these two lines. Any reason?

ister.py
```diff
@@ -447,6 +455,21 @@ def copy_os(args, template, target_dir):
     if args.fast_install:
         run_command("rm -rf {0}".format(args.statedir))
 
+
+def copy_os_dnf(args, template, target_dir):
```
bryteise: Needs unit test

ister.py
```diff
@@ -412,9 +412,17 @@ def get_current_format():
...
-    add_bundles(template, target_dir)
+    package_manager = template["SoftwareManager"]
+    LOG.info("Starting {0}. May take several minutes".format(package_manager))
+    if package_manager == "swupd":
+        copy_os_swupd(args, template, target_dir)
+    elif package_manager == "dnf":
+        copy_os_dnf(args, template, target_dir)
+
 
+def copy_os_swupd(args, template, target_dir):
```
bryteise: Needs unit test.

ister.py
```diff
@@ -412,9 +412,17 @@ def get_current_format():
 def copy_os(args, template, target_dir):
     """Wrapper for running install command
     """
-    LOG.info("Starting swupd. May take several minutes")
-    add_bundles(template, target_dir)
+    package_manager = template["SoftwareManager"]
```
bryteise: Should update unit test to call out specificly running swupd and add test for dnf.

ister.py
gtkramer: Only once, just for swupd

ister.py
gtkramer: Ah, ok.  I'll take this out in the "Remove legacy definitions" commit

ister.py
gtkramer: I split out my changes a little poorly across the commit, but the order was important to get the tests to pass because the command string changes otherwise

ister.py
```diff
@@ -962,6 +989,14 @@ def validate_version_template(template):
         raise Exception("Invalid version string (must be 'latest')")
 
 
+def validate_softmgr_template(template):
```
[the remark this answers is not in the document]
gtkramer: Ah, thank you.  I'll add a test for this.

ister.py
gtkramer: I just added tests for the dnf command generation.

ister.py
gtkramer: Could the existing unit tests be used here because we're defaulting to swupd? Or, maybe I should add a unit test for checking the new copy_os function works to choose the correct copy_os_xxx as expected?

bryteise: The latter, we want to ensure the right one is picked, can stub out the actual call to copy_os_xxx in the test.

gtkramer:
I've gone through and cleaned up all of the commits.  I think they now more logically group changes while minimizing changes done.  I've also added more tests to help with closing the testing gap.  Things look ready on my end.  Can you take another look?

ister_test.py
```diff
@@ -1321,20 +1321,6 @@ def add_bundles_good():
     commands_compare_helper(commands)
 
 
-@open_wrapper("good", "1\n")
-def get_current_format_good():
```
bryteise: I expect the test needs to be removed from the tests array in this commit too.

ister_test.py
```diff
@@ -3126,7 +3126,23 @@ def validate_template_bad_version():
...
+def validate_version_template_bad_version_str():
+    """Bad validate_version_template version string"""
+    exception_flag = False
+    template = {"PartitionLayout": [],
+                "FilesystemTypes": [],
+                "DestinationType": [],
+                "PartitionMountPoints": [],
+                "Version": "20110"}
+    try:
+        ister.validate_template(template)
```
bryteise: I expect this and the num test to call the validate_version_template function directly.

ister.py
```diff
@@ -428,16 +416,36 @@ def copy_os(args, template, target_dir):
...
+        cmd += " --certpath={0}".format(args.cert_file)
+    if shutil.which("stdbuf"):
+        cmd = "stdbuf -o 0 {0}".format(cmd)
+    cmd_env = get_cmd_env(template)
+    run_command(cmd, environ=cmd_env, show_output=True)
 
     if args.fast_install:
         run_command("rm -rf {0}".format(args.statedir))
 
+def get_cmd_env(template):
```
bryteise: Ah this one slipped by me but it this should have a test function too.

ister_test.py
```diff
@@ -2023,7 +2089,7 @@ def get_template_good():
     try:
         template = ister.get_template("file://{0}/test.json"
                                       .format(os.getcwd()))
-        if template != {"test": 1}:
+        if template != {"test": 1, "SoftwareManager": "swupd"}:
```
bryteise: Mind adding another test here for dnf?

bryteise:
Almost there just a couple minor tweaks and I think we can merge it.

ister_test.py
gtkramer: This tests whether when missing a SoftwareManager key that it gets added and defaults to swupd.  Not sure if it makes sense to have a dnf test for this?

ister.py
gtkramer: Ahh, even this?  Ok  :)

ister_test.py
gtkramer: Oops, missed this.  I'll make the change.

ister_test.py
gtkramer: It's already taken out ;)

ister_test.py
bryteise: Ah sorry, I was thinking of tests for validating that dnf and swupd are okay to use.

ister_test.py
bryteise: Ah, okay sorry I didn't see that X(.

ister_test.py
gtkramer: I think I might have that covered with copy_os_switch_xxx (dnf and swupd)?

gtkramer:
I accidentally squashed the commit that did prep work for enabling dnf X( during a rebase.  I decided to squash one other commit to group work at a little bit of a higher level.  Should I re-separate anything?

I went through and tidied up a couple of little things for consistency with other tests and their description terminology.  I just ran through another round of testing and things still look good after the updates.

gtkramer:
@mdhorn are you doing a rewrite of the installer?  Where should I go to work on adding dnf support to the new installer?

ister_test.py
bryteise: Ah yes indeed.

PR #132, opened 2018-05-16
ister: integration of encryption changes and testing

josuedhg:
integration of root encryption in ister
ister.py:
change the place for assignation of the Software Manager default in order
to use ister.py in an import instead of call it as a script

ister_gui.py:
FIX a wrong bundle name to add an initrd for root encryption

ister_test.py:
modify tests to match with the ister.py changes

testing:
validate_release.py:
Add validation for root encryption process

ister-encryption.service:
service ta call encryption.expect and test the encryption installation
process

encryption.expect:
script to automate encryption installation process

post-encryption.expect:
script to verify that encryption process produce a bootable installation

Signed-off-by: Josue David Hernandez <josue.d.hernandez.gutierrez@intel.com>


ister.py
```diff
@@ -1012,6 +1009,10 @@ def validate_version_template(template):
 def validate_softmgr_template(template):
     """Attempt to verify the package manager is sane
     """
+    # Supply default SoftwareManager value if not defined for backwards compatibility
```
bryteise: So the placement here I agree is a little confusing but as discussed in the PR where this was added it does make it easier to avoid modifying the template after it was read it (I realize we do have an exception for this in the code but that can't be done early unlike this one). Perhaps if we get more cases like this a set defaults function would make sense but for now I'm okay with where this code was.

ister_gui.py
```diff
@@ -3271,7 +3271,10 @@ def automatic_install(self):
         args = ister.handle_options(ister_cmd)
 
         ister.LOG = ister.logging.getLogger('ister')
-        ister.handle_logging(args.loglevel, ister_log, AlertLoggerHandler)
+        try:
+            ister.handle_logging(args.loglevel, ister_log, AlertLoggerHandler(title))
+        except Exception as e:
+            self.logger.debug(e)
```
bryteise: This seems like it wouldn't do much if the logger can't be setup. Also pretty unrelated to the change over all.

validate_release.py
```diff
@@ -63,30 +63,39 @@ def validate_installer(args, expectfile, unitfile):
                          '-bios', args.bios])
     cp.check_returncode()
 
-    print(">>> Installing boot canary into installer-target.mg")
-    cp = subprocess.run(['sudo', './install-canary.sh', 'installer-target.img'])
-    cp.check_returncode()
+    if not alternative_check:
```
bryteise: This is a bit messy but I think fine for now, we'll see if we get more of these to consolidate.

post-encryption.expect
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/expect -f
+set ovmf [lindex $argv 0]
+spawn sudo qemu-system-x86_64 -enable-kvm -nographic -m 1024 -cpu host -drive file=installer-target.img,if=virtio,aio=threads -net nic,model=virtio -net user,hostfwd=tcp::2233-:22 -smp 2 -bios $ovmf
```
bryteise: I'm not super in favor of doing this in the expect script but this isn't particularly problematic for now, we can come back to this another time.

ister_gui.py
bryteise: I mean I don't know, I imagine that if somebody gave a logfile location they couldn't write to that it may be better to just fail rather than not log. I'm open to ideas.

josuedhg: mmm ok, changing it

ister.py
josuedhg: ok, we can improve this in the future

validate_release.py
josuedhg: ok

post-encryption.expect
josuedhg: I don't see another way to test the root encrypted password during boot time, if you think in a way we could change in a future update

josuedhg:
can we merge it?

josuedhg:
this depend on this commit to be released: https://github.com/clearlinux/clr-boot-manager/commit/2aba4168f86eea6f15b1ed40e7aa8dcda7875c7d

ister_gui.py
```diff
@@ -3271,9 +3271,9 @@ def automatic_install(self):
         args = ister.handle_options(ister_cmd)
 
         ister.LOG = ister.logging.getLogger('ister')
-        ister.handle_logging(args.loglevel, ister_log, AlertLoggerHandler)
 
         try:
+            ister.handle_logging(args.loglevel, ister_log, AlertLoggerHandler(title))
```
bryteise: This change is unrelated to the rest of the commit. It should be in its own commit.

ister.py
bryteise: Um I don't see why this part of the change needs to be kept. The explanation in the git commit doesn't make sense to me either as ister is imported successfully already in the gui and test scripts.

ister_gui.py
josuedhg: ok

josuedhg: ready

ister.py
josuedhg: the changes to encryption needs to keep the password hide, if ister.py is called in a subprocess this will be store in the template.json in the file system, the way to avoid this was change how ister_gui.py calls ister.py instead use a subprocess ister.py is used as python module and its pass the template dict directly to its functions, https://github.com/bryteise/ister/pull/125 the previous place where was this default assignation is where the json is being validated and that function is executed just when the ister.py is called directly in the command line

bryteise: I think the correct change here is to update ister_gui.py to include the SoftwareManager in the template then not move where the default is added. That can be done in another commit so you don't need to rebase.

josuedhg: if it is an optional parameter, even if it is not defined it most be assigned, if it is not moved you force to define this option when ister is used as a module

bryteise: My thinking is that install_os requires a valid in memory template given to it. A template file can have an optional component but not that isn't what install_os takes as an argument. I don't want to have any function in the module able to handle an incomplete template.

josuedhg: ok, changing it

josuedhg: Ready

PR #135, opened 2018-07-12
Adding custom mirror URL fields to installer

mdpitterle:
Summary of changes:

- set argument default to None when cmdline value not specified (no hardcoding)
- allow user to optionally modify values independently for the content or version URL's to alternate mirrors
- if there is a URL modification, or a command line value that verify as 'good', ister will write the values to the target OS

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -453,10 +453,18 @@ def copy_os_swupd(args, template, target_dir):
     cmd = "swupd verify --install"
     cmd += " --path={0}".format(target_dir)
     cmd += " --manifest={0}".format(template["Version"])
-    if args.contenturl:
-        cmd += " --contenturl={0}".format(args.contenturl)
-    if args.versionurl:
-        cmd += " --versionurl={0}".format(args.versionurl)
+    if args.contenturl or template.get("MirrorURL"):
+        if template.get("MirrorURL"):
+            cmd_mirror_url =  template.get("MirrorURL")
+        else:
+            cmd_mirror_url =  args.contenturl
+        cmd += " --contenturl={0}".format(cmd_mirror_url)
+    if args.versionurl or template.get("VersionURL"):
+        if template.get("VersionURL"):
+            cmd_version_url =  template.get("VersionURL")
+        else:
+            cmd_version_url =  args.versionurl
+        cmd += " --versionurl={0}".format(cmd_version_url)
     if args.format:
         cmd += " --format={0}".format(args.format)
     cmd += " --statedir={0}".format(args.statedir)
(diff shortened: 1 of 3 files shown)


ister.py
```diff
@@ -664,6 +674,30 @@ def set_hostname(template, target_dir):
...
+    LOG.info("Setting custom mirror url")
+    path = '{0}/etc/swupd/'.format(target_dir)
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+    with open(path + "mirror_contenturl", "w") as file:
+        file.write(target_mirror_url)
+
+
+def set_version_url(target_version_url, target_dir):
```
bryteise: Please call out that this is a mirror version url rather than the actual version url (since that exists as a file too).

ister_gui.py
```diff
@@ -1676,20 +1726,32 @@ def handler(self, config):
         config = self.config
         return self._action
 
+
```
bryteise: Extra whitespace here that should be cleaned up.

ister.py
```diff
@@ -453,10 +453,18 @@ def copy_os_swupd(args, template, target_dir):
     cmd = "swupd verify --install"
     cmd += " --path={0}".format(target_dir)
     cmd += " --manifest={0}".format(template["Version"])
-    if args.contenturl:
-        cmd += " --contenturl={0}".format(args.contenturl)
-    if args.versionurl:
-        cmd += " --versionurl={0}".format(args.versionurl)
+    if args.contenturl or template.get("MirrorURL"):
+        if template.get("MirrorURL"):
```
bryteise: Ah sorry, I missed one thing because of my thinking that we wouldn't end up using the mirrorURL in the template. I think the order should be switched. Use the args if it exists otherwise use the template. The commandline for ister.py should supercede. This is an odd case though.

ister.py
```diff
@@ -1422,6 +1478,21 @@ def install_os(args, template):
         copy_os(args, template, target_dir)
         add_users(template, target_dir)
         set_hostname(template, target_dir)
+        # UI changes override command line
+        if args.contenturl or template.get("MirrorURL"):
```
bryteise: I think that we shouldn't persist the args to disk, only the template in this case.

ister.py
mdpitterle: If ister.py is used directly, the template value will never be set anyway - so no harm, and the code will do the right thing here.

ister.py
mdpitterle: @mdhorn - isn't it true that the new installer will persist the command line args to disk on target host?   My intent here is to be consistent with what new installer will do (physical servers only)

ister.py
```diff
@@ -453,10 +453,18 @@ def copy_os_swupd(args, template, target_dir):
     cmd = "swupd verify --install"
     cmd += " --path={0}".format(target_dir)
     cmd += " --manifest={0}".format(template["Version"])
-    if args.contenturl:
-        cmd += " --contenturl={0}".format(args.contenturl)
-    if args.versionurl:
-        cmd += " --versionurl={0}".format(args.versionurl)
+    if args.contenturl or template.get("MirrorURL"):
```
bryteise: So this is for the swupd command, it should only look at the arg.contenturl

ister.py
```diff
@@ -1422,6 +1478,12 @@ def install_os(args, template):
         copy_os(args, template, target_dir)
         add_users(template, target_dir)
         set_hostname(template, target_dir)
+        if template.get("MirrorURL"):
+            cmd_mirror_url =  template.get("MirrorURL")
```
bryteise: Minor nit, extra space after =. What do you think about just passing template instead of the specific thing from the template? All the other calls around this one do that so it would be more uniform.

ister.py
mdpitterle: Well, one of the main points of this feature (when using the GUI) is to allow the user to modify where to get content during the build (for testing, or to improve build speed on a local mirror perhaps) - if I don't honor this on swupd cmd line - it will likely be confusing to the user.

ister.py
mdpitterle: No problem...now that I am not writing args to any destination - I don't need the extra logic, and can make it consistent like the other calls

ister.py
bryteise: My thought is to have the gui update the args passed to ister as well as the template rather than have ister use the args or the template.

PR #136, opened 2018-08-03
A couple of useful options

dedekind:
Make it possible for the user to tweak the resulting installation before rebooting into ClearLinux.


bryteise:
Closing in favor of #139

PR #140, opened 2018-08-15
Improve isterconf

dedekind:


bryteise:
Seems reasonable to speculatively try and load a template in this case.

PR #141, opened 2018-08-15
Add labels support

dedekind:


bryteise:
Nice change, happy to have label support!

PR #142, opened 2018-08-21
ister_gui: Fix bug during Bundles initialization

mbelluzzo:
The list is populate during the creation of the installation object with
the bundle list provided by the template found at /etc/ister.json,
inject there during the manufacturing of the installer image. Just after
that is was being cleared and therefore the templates list was never
respected.

That initialization is enough for the code to properly work since at
that point the code is executed only once, no matter if the user rolled
back some screens.

Signed-off-by: Murilo Belluzzo <murilo.belluzzo@intel.com>

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -3323,9 +3323,6 @@ class Installation(object):
         """Starts up the installer ui"""
         pprint.PrettyPrinter(indent=4)
         step = self.start
-        # initiate the bundles list at the start, not in a screen, so it does
-        # not get overwritten when a user returns to that screen.
-        self.installation_d['Bundles'] = list()
         i = 0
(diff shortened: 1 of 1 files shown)


bryteise:
I need to poke this a bit more, still a little fuzzy on what was going on in this code =/.

bryteise:
So the template's bundle list from ister.json in the installer image wasn't intended to be used as content for the gui and prepopulate the bundle selection list.

mbelluzzo:
Is not content for the gui, is bundles that must be installed no matter
what

bryteise:
So I agree with the idea of how we end up using the template in the gui but as the gui is getting replaced soon with clr-installer I'm inclined to keep the code as is.

PR #143, opened 2018-08-27
Run command cleanup

dedekind:
A couple of cleanups. No functional changes.


ister.py
```diff
@@ -87,30 +87,22 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None,
...
+                output.append(decoded_line)
+                if show_output:
+                    LOG.info(decoded_line)
+                elif log_output:
+                    LOG.debug(decoded_line)
 
         if proc.poll() and raise_exception:
-            if stderr:
-                LOG.debug("Error {0}".format('\n'.join(stderr)))
+            if outputs[1]:
```
bryteise: I don't see where outputs is ever getting assigned outside of the initial empty list.

ister.py
```diff
@@ -87,30 +87,22 @@ def run_command(cmd, raise_exception=True, log_output=True, environ=None,
...
+        outputs = []
+        for lines in (proc.stdout, proc.stderr):
+            output = []
+            for line in lines:
+                decoded_line = line.decode('ascii', 'ignore').rstrip()
+                output.append(decoded_line)
+                if show_output:
+                    LOG.info(decoded_line)
+                elif log_output:
+                    LOG.debug(decoded_line)
```
bryteise: I think you wanted outputs.append(output) after the inner for loop.

dedekind: Sorry, I fixed this all yesterday and tested this patch as well. But I did not push out and you ended up looking at the initial version.

dedekind:
Hi William, I initially sent you old version of my patches. But after this I pushed the right patches which I actually tested and use today.

bryteise:
Ah okay, I'll relook at it today then.

bryteise:
Sorry, took a little longer to get to review, looks good. Thanks!

PR #146, opened 2018-09-18
Allow specifying authentication keys by value

dedekind:
Improve ister and allow specifying the SSH key by value as well.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -52,6 +52,8 @@ import subprocess
 import sys
 import tempfile
 import time
+import base64
+import binascii
 import traceback
 import urllib.request as request
 from urllib.error import URLError, HTTPError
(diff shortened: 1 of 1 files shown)


ister.py
```diff
@@ -1113,8 +1113,13 @@ def validate_user_template(users):
                                 .format(user))
 
         if key:
-            with open(user["key"], "r") as key_file:
-                user["key"] = key_file.read()
+            try:
+                with open(user["key"], "r") as key_file:
+                    user["key"] = key_file.read()
+            except FileNotFoundError:
```
bryteise: Having this be an exact error is good but what do you think of a general except case for problems with the file read?

dedekind: Well, the code did not handle the general OSError so I did not do it either. Kind of preserved old behavior. The old behavior was - any error reading the file - raise an exception.

If we cannot read the file for other reasons (permissions, or it is not a file, etc), the exception will be raised and caught in main(). The error message will be printed. For file I/O exceptions this typically includes file name. When debugging is enabled - we'll see also the stacktrace.

I do not think we should fall back to assuming this is an inline key for any exception. I think anything but FileNotFoundError is really a problem ans we should promptly fail rather than continue and do something users do not expect. If in the future we have users who have an issue with this, we can adjust the behavior. At this point it is good enough in my opinion.

Let me know if there is something specific in your mind.

dedekind:
I'll send an updated pull request. I'll add a small sanity check to verify if the contents looks like an inline ssh key. This will fix the failing test.

ister.py
bryteise: Ah right, sorry for the noise as I clearly temporarily forgot how python handles exceptions X(.

PR #148, opened 2018-09-20
Changing root user

dedekind:
This pull requests makes it possible to change the 'root' user via 'ister.json', for example change root password and install root user SSH key. On top of this, the pull request fixes few bugs and introduces few improvements.

My use-case is the following. I have a SUT (System Under Test) in a fully isolated environment. I want my fully automated and unattended installation to set a pre-defined root password, root SSH key, and enable root SSH login. As simple as this. And I do not want to bring up 'cloud-init' server just for this simple purpose.

Here is the ister.json template that I use. It only works with patches introduced in this pull request. The ister.json below is acutally a Jinja2 template, which I render to get the final 'ister.json'. But this is an unimportant detail and I prefer shoing the Jinja2 template because it is more readable than the final result.

```
{
    "DestinationType" : "physical",
    "PartitionLayout" :      [{"disk" : "sda", "partition" : 1, "size"  : "1G",   "type"  : "EFI"},
                              {"disk" : "sda", "partition" : 2, "size"  : "rest", "type"  : "linux"}],
    "FilesystemTypes" :      [{"disk" : "sda", "partition" : 1, "type"  : "vfat", "label" : "boot"},
                              {"disk" : "sda", "partition" : 2, "type"  : "ext4", "label" : "root"}],
    "PartitionMountPoints" : [{"disk" : "sda", "partition" : 1, "mount" : "/boot"},
                              {"disk" : "sda", "partition" : 2, "mount" : "/"}],
    "Hostname"   : "{{ host }}",
    "PostChrootShell" : [{{ "echo 'PermitRootLogin yes' >> /etc/ssh/sshd_config" | tojson}},
                         {{ "echo 'Subsystem sftp /usr/libexec/openssh/sftp-server' >> /etc/ssh/sshd_config" | tojson}}],
    "Users"  :  [{"username" : "root",
                  "password" : {{ rootpass_hash | tojson}},
                  "key" : {{ authkey | tojson }}}],
    "Bundles" : ["kernel-native", "user-basic"],
    "Version" : "latest"
}
```

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -530,6 +530,7 @@ class ChrootOpen(object):
         try:
             self.old_root = os.open("/", os.O_RDONLY)
             os.chroot(self.target_dir)
+            os.chdir("/")
         except Exception:
             raise Exception("Unable to setup chroot to create users")
 
--- a/ister.py
+++ b/ister.py
@@ -548,6 +548,12 @@ class ChrootOpen(object):
             raise Exception("Unable to restore real root after chroot")
 
 
+def get_user_homedir(username):
+    """Returns user's home directory path."""
+    if username == "root":
+        return os.path.join(os.sep, "root")
+    return os.path.join(os.sep, "home", username)
+
 def create_account(user, target_dir):
     """Add user to the system
 
@@ -555,7 +561,7 @@ def create_account(user, target_dir):
     passwordless login. Also add a new group with same name as the user
     """
 
-    os.makedirs(target_dir + "/home", exist_ok=True)
+    os.makedirs(target_dir + get_user_homedir(user["username"]), exist_ok=True)
     if user.get("uid"):
         command = "useradd -U -m -u {0} {1}"\
             .format(user["uid"], user["username"])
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -562,18 +562,27 @@ def create_account(user, target_dir):
...
 
     os.makedirs(target_dir + get_user_homedir(user["username"]), exist_ok=True)
-    if user.get("uid"):
-        command = "useradd -U -m -u {0} {1}"\
-            .format(user["uid"], user["username"])
+    if user.get("password"):
+        opts = "-p '{0}'".format(user["password"])
     else:
-        command = "useradd -U -m {}".format(user["username"])
+        opts = "-p ''"
```
bryteise: This will allow users without a password set to login by default which is not something we want to allow. The option should be an empty string and if the behavior is desired an empty string should be used as the password in the template.

ister.py
```diff
@@ -562,18 +562,27 @@ def create_account(user, target_dir):
...
 
-    command += (" -p {0}".format(user["password"])
-                if user.get("password") else " -p ''")
+    command = "useradd -U -m {0} {1}".format(opts, user["username"])
 
     with ChrootOpen(target_dir) as _:
-        run_command(command)
-
+        _, stderr, ret = run_command(command, raise_exception=False)
+        if ret == 9:
```
bryteise: The commit message has a typo saying 0 instead of 9.

ister.py
bryteise: Ugh sorry, that was how it has been X(. Well must have had a good reason for that and it passed security review so not changing the behavior now.

dedekind:
OK, I've fixed the commit message (0->9) and added the testcase.

PR #150, opened 2018-09-21
Remove 15 secs sleep

dedekind:
This patch speeds up ister installer by removing the 15 seconds unconditional sleep. I understand why this sleep was added, but this is a very crude solution for the problem. This pull request offers a nicer solution - instead of sleeping for 15 seconds we keep trying with 15 seconds timeout.

This pull request also implements several cleanups resulting in a bit more structured and readable code.

This pull request includes the 3 previous pull requests. It is based on top of the "enhance run_command" work. I'll rebase it if/when needed. Comparing to the "enhance run_command" pull request, this adds the following patches:

Remove useless os.unlink
Smarter ister.conf download
Generalise 'check_kernel_cmdline()'
Simplify check_kernel_cmdline

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -1386,7 +1386,31 @@ def validate_template(template):
     LOG.debug(template)
 
 
-def process_kernel_cmdline(f_kcmdline, sleep_time=15):
+def download_ister_conf(uri, timeout=15):
+    """Download the ister.conf/ister.json file from 'uri' to a local temporary
+    file and return the temporary file path. The timeout argument specifies for
+    how long to try downloading the file."""
+
+    tmpfd, abs_path = tempfile.mkstemp()
+    LOG.debug("ister_conf tmp file = {0}".format(abs_path))
+
+    start_time = time.time()
+    while True:
+        try:
+            with request.urlopen(uri) as response:
+                with closing(os.fdopen(tmpfd, "wb")) as out_file:
+                    shutil.copyfileobj(response, out_file)
+                    return abs_path
+            os.unlink(abs_path)
+        except Exception as err:
+            # In a PXE environment it's possible systemd launched us before the
+            # network is up. Therefore, keep trying for 'timeout' seconds.
+            if time.time() - start_time > timeout:
+                raise Exception("failed to download ister.conf from '{0}': {1}"\
+                                .format(uri, err))
+
+
+def process_kernel_cmdline(f_kcmdline):
     """Some ister options can be passed via carnel configuration file. For
     example, 'isterconf=<path>' can be used for passing ister configuration file
     (AKA 'ister.conf') or ister template file (AKA 'ister.json'). This function
@@ -1406,20 +1430,8 @@ def process_kernel_cmdline(f_kcmdline, sleep_time=15):
             ister_conf_uri = opt.split("=")[1]
 
     LOG.debug("ister_conf_uri = {0}".format(ister_conf_uri))
-
-    # Fetch the file
     if ister_conf_uri:
-        tmpfd, abs_path = tempfile.mkstemp()
-        LOG.debug("ister_conf tmp file = {0}".format(abs_path))
-        # in a PXE environment it's possible systemd launched us
-        # before the network is up. This is primitive but effective.
-        # And generally only pxe boots will trigger this.
-        time.sleep(sleep_time)
-        with request.urlopen(ister_conf_uri) as response:
-            with closing(os.fdopen(tmpfd, "wb")) as out_file:
-                shutil.copyfileobj(response, out_file)
-                return abs_path
-        os.unlink(abs_path)
+        return download_ister_conf(ister_conf_uri)
     return None
 
 
--- a/ister.py
+++ b/ister.py
@@ -1401,7 +1401,6 @@ def download_ister_conf(uri, timeout=15):
                 with closing(os.fdopen(tmpfd, "wb")) as out_file:
                     shutil.copyfileobj(response, out_file)
                     return abs_path
-            os.unlink(abs_path)
         except Exception as err:
             # In a PXE environment it's possible systemd launched us before the
             # network is up. Therefore, keep trying for 'timeout' seconds.
--- a/ister.py
+++ b/ister.py
@@ -1386,10 +1386,14 @@ def validate_template(template):
     LOG.debug(template)
 
 
-def check_kernel_cmdline(f_kcmdline, sleep_time=15):
-    """Check if ister configuration (AKA ister.conf) is defined via kernel
-    command line (PXE environments). If it is defined, then download it and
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -66,6 +66,90 @@
...
+            if not data:
+                continue
+            data, partial = extract_full_lines(partial + data)
+            if data:
+                info["queue"].put((streamid, data.strip()))
+    except BaseException as err:
+        LOG.error(err)
+
+    if partial:
+        info["queue"].put((streamid, partial))
```
bryteise: I don't think it matters much but I'm wondering if we should strip partial even though it won't have the trailing newline.

bryteise: Hrm looking at your next commit you take out strip on data so I guess that answers my question. I'll raise it on that commit.

ister.py
```diff
@@ -66,6 +66,90 @@
...
+
+    full, partial = ("", "")
+    for line_match in re.finditer("(.*\n)|(.+$)", text):
+        if line_match.group(2):
+            partial = line_match.group(2)
+            break
+        full += line_match.group(1)
+    return (full, partial)
+
+def stream_fetcher(info, streamid):
```
bryteise: This and extract_full_lines were unused and needed imports that weren't added at the time. I think I reviewed it correctly but I'm not sure if this was intended to be included here or not.

ister.py
```diff
@@ -124,26 +130,41 @@ def stream_fetcher(info, streamid):
...
+                                            args=(info, streamid)))
+            threads[-1].start()
+
+    output = ([], [])
+    try:
+        while True:
+            streamid, line = info["queue"].get()
+            if line is not None:
+                # Drop single newline.
+                line = line[:-1]
```
bryteise: So instead of stripping you are only removing the newline which changes the output again I'm not super worried about it but extra whitespace seems odd to be adding to logs.

bryteise: Also this will drop the last character of the partial line added at the end. I think using strip here is probably best.

ister_test.py
```diff
@@ -4136,7 +4109,6 @@ def mock_os_unlink(path):
     finally:
         tempfile.mkstemp = mkstemp_orig
         shutil.copyfileobj = cfo_orig
-        os.unlink = unlink_orig
```
bryteise: Ugh this just makes me note that this was very untested as the COMMAND_RESULTS isn't being verified X(. Not your problem though.

bryteise:
Okay review complete. The run_command change is a huge improvement, thanks!

ister.py
dedekind: I'll check and if this is the case I'll amend the patches and will make sure that every patch adds the import it requires. I was doing a lot of rebase -i and could make a mistake when splitting patches.

ister.py
dedekind: I'll amend the patches to make them more logical. I first have the .strip() but then realized it is incorrect, because it will strip _all_ the newlines - more than it should.

ister.py
dedekind: I had a strip and then removed it because I believe it removes more than needed. I'll re-examine.

ister.py
```diff
@@ -1387,7 +1387,32 @@ def validate_template(template):
...
+            with request.urlopen(uri) as response:
+                with closing(os.fdopen(tmpfd, "wb")) as out_file:
+                    shutil.copyfileobj(response, out_file)
+                    return abs_path
+            os.unlink(abs_path)
+        except Exception as err:
+            # In a PXE environment it's possible systemd launched us before the
+            # network is up. Therefore, keep trying for 'timeout' seconds.
+            if time.time() - start_time < timeout:
+                pass
```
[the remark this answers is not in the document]
dedekind: Well, this is the same thing - we are in the exept block and 'pass' will just take us out of the exception block and we'll start the new loop iteration. May be 'continue' is a bit more readable, but in my eye there is no difference. I can put 'continue' there.

dedekind:
Thanks for review. I'll address your comments and re-submit.

ister.py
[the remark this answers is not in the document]
bryteise: Hrm from the documentation and testing I just went through I'm not sure about pass working the way you describe. Pass is documented as a noop and so it would only impact the if statement because python wants an indentation. The raise Exception following the if statement would still get run.

Maybe my little toy example below isn't quite correct though and I'd be interested in learning what my misunderstanding is (maybe I just shouldn't review on Saturday morning X( ).

`def foo():
    try:
        i = 0
        raise Exception("foo")
    except Exception as err:
        if i == 0:
            pass
        raise Exception("bar")`

And if I run this it raises an exception "bar".

ister.py
dedekind: Initially I had one commit, but then I split it to make review easier, but made a mistake. I've fixed this now.

ister.py
dedekind: OK, the problem was that extract_full_lines returned all the full lines as one string. Not as a list of lines. So it could return "text\n\n\n", for example, because the command printed three newlines in a row. Using .strip() would remove all the three, and subsequent 'LOG.debug()' would add one. So the end result is that newlines that must have been there would be removed.

I've changed this and now extract_full_lines returns a per-line list, already stripped, so nothing else should be done to the output.

ister.py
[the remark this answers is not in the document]
dedekind: **blush**
You are correct. Will fix, thanks.

ister.py
```diff
@@ -1386,7 +1386,31 @@ def validate_template(template):
...
+    LOG.debug("ister_conf tmp file = {0}".format(abs_path))
+
+    start_time = time.time()
+    while True:
+        try:
+            with request.urlopen(uri) as response:
+                with closing(os.fdopen(tmpfd, "wb")) as out_file:
+                    shutil.copyfileobj(response, out_file)
+                    return abs_path
+            os.unlink(abs_path)
```
bryteise: I think the previous version had this unlink removed because it couldn't be executed (either exception or success). Was there a reason for adding this back or am I confused because of my bad merging?

ister_test.py
```diff
@@ -4148,7 +4148,7 @@ def mock_os_unlink(path):
     os.unlink = mock_os_unlink
 
     try:
-        ister.process_kernel_cmdline("foo")
+        ister.process_kernel_cmdline("foo", timeout=0.1)
```
bryteise: Since the structure of the download changed I think this test probably should be dropped as there is just a single exception block for the logic.

dedekind:
Updated the pull request:
* Returned the accidentally dropped "unlink" removal patch, thanks for noticing this.
* I've removed 2 testcases as you suggested.
Thanks!

bryteise:
Thanks for this, nice improvements!

PR #151, opened 2018-09-24
No default empty password

dedekind:
This pull request is on top of the "Remove 15 secs sleep". Please, review/pull this request after "Remove 15 secs sleep".

This pull request addresses the issue you spotted while reviewing one of my previous pull requests: if user password is not specified, we create the user with an empty password, and this is poor choice for the default password from security point of view. This is not something I introduced, this behavior existed for long time, and I hope changing it won't break users. But it is better to change now than later.

With this pull request if the password was not specify, we just create the user without any password and using this user to log in will be prohibited. Compare this to the old behavior, where we use empty password and anyone can login.

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -648,22 +648,23 @@ def create_account(user, target_dir):
     """
 
     os.makedirs(target_dir + get_user_homedir(user["username"]), exist_ok=True)
-    if user.get("password"):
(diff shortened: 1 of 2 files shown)


bryteise:
Great improvement, thanks!

bryteise:
Merged outside of github due to my stupid merging earlier.

PR #152, opened 2018-09-24
New cmdline opts

dedekind:
This pull request is on top of "No default empty password" pull request. Please, review/pull this one after "No default empty password".

This pull request adds a way of using the ister-based ClearLinux PXE images in fully automated environments. The details including the usage scenario are described in the commit messages.

There is one outstanding problem though. In ClearLinux ister is run by the following systemd service:

``
cat /usr/lib/systemd/system/multi-user.target.wants/ister-provision.service
[Unit]
Description=ister graphical installer

[Service]
Type=oneshot
ExecStart=/usr/bin/python3 /usr/bin/ister.py
ExecStartPost=/usr/sbin/reboot

[Install]
WantedBy=multi-user.target
``

Notice "ExecStartPost" - as soon as ister exists, the system is rebooted. Therefore, I cannot exit ister on the 'ister.exit' cmdline option and I just make it sleep forever, and user just runs a new instance.

I plan to improve ister and teach it to do the reboot at the end on success. In case of failure it is better not to reboot and let the user be able to investigate the problem. Then I'll try to work with the ClearLinux team to remove that "ExecStartPost=/usr/sbin/reboot". And then we can make ister truly exit on 'ister.exit'.

Let me know what you think about all this. I can only say that I am using this approach in my environment to overcome super-strict security limitation in the ClearLinux PXE installer image (no root login, must set root password from serial, no root SSH).

Thanks!


ister.py
```diff
@@ -1410,6 +1410,33 @@ def download_ister_conf(uri, timeout=15):
...
+    sshdir = os.path.join("etc", "ssh")
+    os.makedirs(sshdir, mode=0o0755, exist_ok=True)
+    with open(os.path.join(sshdir, "sshd_config"), "a") as fobj:
+        os.fchmod(fobj.fileno(), 0o600)
+        fobj.write("# Added by ister due to 'ister.root_ssh'.\n"
+                   "PermitRootLogin yes\n"
+                   "PermitEmptyPasswords yes\n"
+                   "Subsystem sftp /usr/libexec/sftp-server\n")
+    # Restart the SSH service.
+    run_command("systemctl restart sshd")
```
bryteise: Mind me poking this over to use sshd.socket rather than the service (I have a few conflicts I am resolving so I can do this when I push those)? Since we generally recommend using the socket and so using the service here might conflict with somebody else who built their own installation own image.

dedekind: Not 100% sure I understood the request correctly, but I'll change this to 'systemctl restart sshd.socket'.

dedekind:
This is updated pull request. It uses sshd.socket.

I also had to add 3 small patches - somehow these modifications were lost, they were part of the previous pull requests, but it looks like you merged an older version of pull request and changes some patches or pieces of it. Not sure what happened.

From my side it was that between the different versions of my pull requests I added these small fixes. Resending them now. Let me know if you want them separate.

dedekind:
Hmm, hold on, this systemctl restart sshd.socket does not work for me ...

dedekind:
It looks like 'restart' does not work the way I though. But this seem to work:
systemctl enable --now sshd.socket

PR #154, opened 2018-10-01
Forcemu

dedekind:
This pull requests fixes mount units creation and enhances this functionality with the 'forcemu' ister.json option.


ister.py
```diff
@@ -503,14 +504,20 @@ def create_mount_unit(filename, uuid, mount, fs_type):
...
-                create_mount_unit(os.path.join(units_dir, filename),
-                                  get_uuid(pnum, base_dev),
-                                  part["mount"], fs_type)
+
+        # Create mount units for the partitions, except for those having standard
+        # GPT type GUIDs, because the standard systemd 'systemd-gpt-auto-generator'
+        # tool will generate the mount points.
+        if part["mount"] in ["/", "/boot", "/srv", "/home"]:
+            continue
+        if part["mount"].startswith("/usr"):
```
bryteise: Not your fault but would you mind fixing this case so that /usr-2 isn't skipped as we really only want to verify not mounting over /usr/ parts by default.

bryteise:
The add 'forcemu' option patch should probably also add a check in validate_partition_mounts to make sure it is set to a bool.

bryteise:
The validate_partition_mounts change is the only one required.

dedekind:
I've added 2 more patches on top. Let me know if they do not address your requests. Thanks!

bryteise:
Thank you very much!

PR #158, opened 2018-10-10
Add a filter in get_list_of_disks to prevent it to pick up devices like /dev/mmcblk0rpmb

lucabelluccini:
https://lwn.net/Articles/682276/


bryteise:
Thanks a bunch!

PR #159, opened 2018-10-18
Fix the OS naming

mdhorn:
Update product name and text on install screen

Signed-off-by: Mark D Horn <mark.d.horn@intel.com>


bryteise:
python3 validate_release.py -b /usr/share/qemu/OVMF.fd -V 0

PR #160, opened 2018-12-17
Add cmdline parameter to make root mount timeout infinite

josuedhg:
Signed-off-by: Josue David Hernandez Gutierrez <josue.d.hernandez.gutierrez@intel.com>

### DIFF
--- a/ister_gui.py
+++ b/ister_gui.py
@@ -2069,7 +2069,7 @@ class ConfigureCmdline(ProcessStep):
     def handler(self, config):
 
         if not self._ui_widgets:
-            self.build_ui_widgets()
+            self.build_ui_widgets(config)
 
         if not self._ui:
             self.build_ui()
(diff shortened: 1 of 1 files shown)


ister_gui.py
```diff
@@ -2085,14 +2085,17 @@ def handler(self, config):
...
         desc = urwid.Text('console=tty0 console=ttyS0,115200n8 quiet '
                           'init=/usr/lib/systemd/systemd-bootchart '
                           'initcall_debug tsc=reliable no_timer_check '
                           'noreplace-smp kvm-intel.nested=1 '
                           'rootfstype=ext4,btrfs,xfs intel_iommu=igfx_off '
                           'cryptomgr.notests rcupdate.rcu_expedited=1 '
                           'i915.fastboot=1  rw rootwait')
+        for dev_entry in config['PartitionMountPoints']:
+            if 'encryption' in dev_entry:
+                self.cmdline_field.set_edit_text("rootflags=x-systemd.mount-timout=infinity,retry=10000")
```
bryteise: What is retry doing here? All I saw on this was that setting rd.luks.options=timeout=0 rootflags=x-systemd.device-timeout=0 would disable the timeout for entering a password.

mdhorn:
The new installer is using
`rootflags=x-systemd.device-timeout=0
`
Which is the correct solution? Even though we are close to EOL'ing ister_gui, we should be consistent.

josuedhg:
@mdhorn both approach are right, but u are right we must bu consistent I will change it

ister_gui.py
josuedhg: the retry is for mount root, but I will keep it as we have in clr-installer, updating...

PR #162, opened 2019-04-15
Use libblockdev instead of pycryptsetup

lpereira:
Cryptsetup 2.0.6 deprecated pycryptsetup, and it has been removed from
version 2.1.0.  All the functionality needed for ister is available in
libblockdev, so make it use that instead.

This has not been tested.  It's just a RFC.

Signed-off-by: Leandro Pereira <leandro.pereira@intel.com>

### DIFF
--- a/ister.py
+++ b/ister.py
@@ -68,7 +68,9 @@ from urllib.error import URLError, HTTPError
 from urllib.parse import urlparse
 from contextlib import closing
 import netifaces
-import pycryptsetup
+import gi
+gi.require_version("BlockDev", "2.0")
+from gi.repository import BlockDev as bd
 
 LOG = None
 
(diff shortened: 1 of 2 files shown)


ister.py
```diff
@@ -68,7 +68,9 @@
 from urllib.parse import urlparse
 from contextlib import closing
 import netifaces
-import pycryptsetup
+import gi
+gi.require_version("BlockDev", "2.0")
+from gi.repository import BlockDev as bd
```
bryteise, line 73: Oh fun, gobject introspection. Well I guess that's making good use of the gnome stack.

bryteise:
Looks fine to me. Everything seems to make sense. I wonder though if we could fail earlier in case of errors in the two new calls for setting up LUKS? I'm not really bothered by this behavior though as it would still fail eventually.

